Issue #15, opened 2022-02-15
Track roxygen-generated NAMESPACE and documentation in git  


brookslogan:
Installing the package via
```
devtools::install_github("cmu-delphi/delphi-epidata-r", ref="fix-rbuildignore-makefile-typo")
```
produces multiple instances of these warnings:
```
Warning messages:
1: replacing previous import ‘jsonlite::unbox’ by ‘rlang::unbox’ when loading ‘delphi.epidata’ 
2: replacing previous import ‘jsonlite::flatten’ by ‘rlang::flatten’ when loading ‘delphi.epidata’ 
```
and attempts to read the documentation, e.g.,
```
help(epirange)
```
come back with
```
No documentation for ‘epirange’ in specified packages and libraries:
```
while cloning and running `devtools::document()` then `devtools::unload()` then `devtools::install()` appears to resolve the above problems.

While it's not the git way to track auto-generated files, it seems to be the R way.  Tracking the NAMESPACE and .Rd files should simplify and improve the installation experience.

PR #18, opened 2022-03-22
Track current documentation in git

dshemetov:
Addresses #15. Might want to do a documentation pass before merging, but I'm unsure how to preview the docs from a PR. We need to set GH pages to host the docs, maybe?


brookslogan:
Previewing docs can be done by
```
devtools::build_manual()
```
and looking at the pdf, or
```
devtools::install()
help(package="delphi.epidata") # for index of help topics --- this isn't supported when we are using devtools::load_all
[plus cleanup:]
remove.packages("delphi.epidata")
[restart R session]
devtools::load_all(export_all=FALSE) # for help on individual topics, we can use install or load_all
## browse through index)
help(delphi.epidata::covid_hosp_facility)
help(delphi.epidata::covid_hosp_facility_lookup)
[...]
```

brookslogan (approved):
Looks good to me.  The documentation looks pretty bare-bones and might need a few corrections or completions, but I think it makes sense to go ahead and merge anyway. 

For previewing documentation / documentation website easily within a PR, maybe we can check with `covidcast` developers who might already know how.  (Also for how they publish the documentation website for `main` somewhere, if that's not already happening.)  One way to preview the pdf manual would be to track the manual in the repo as well, but that's a bit inconvenient/forgetfullness-prone as devtools::document does not build the pdf manual by default and it takes a bit of time.

PR #39, opened 2022-08-04
Perform package rename from `delphi.epidata` to `epidatr`

brookslogan:
Should be performed after the repo rename has occurred (which is why this is currently marked draft).


dshemetov (approved):
The repo has been renamed from `delphi-epidata-r` to `epidatr` so this should be good to go.

PR #41, opened 2022-08-10
Fix some remaining references to old repo&pkg names

brookslogan:


brookslogan:
Double-checked the changes included here and they seem to be right, and all but one don't seem to have any potential impact on package behavior.  The one that might is the Rproj file renaming; it seems like at this point, this [actually has no impact](https://github.com/rstudio/rstudio/issues/1909), and the repo rename itself would have the impact of renaming the project.  (Perhaps at some point, the project name was instead set [based on](https://community.rstudio.com/t/project-naming-is-inappropriate/51963) the project file's filename?)  So maybe an optional migration step for developers on epidatr is to re-clone or rename the clone dir.  Will see if there's a good place to note this.  Just merging now due to the simplicity.

PR #42, opened 2022-08-10
Apply rename and slightly adjust format of User-Agent

brookslogan:
- Rename `delphi_epidata` -> `epidatr`
- Add slash between package name and version, to match epidatpy's format for
  User-Agent (and the convention for specifying versions in User-Agent)


brookslogan:
Thanks @dshemetov for catching this!

Issue #54, opened 2022-10-19
Complete first testing&documentation pass for remaining endpoints


brookslogan:
We are merging #23 as-is.  There are a few infrequently-used endpoints that have some remaining TODOs from this first pass.  Quoting from the list there:

- [x] dengue_nowcast (need valid locations, but API is lacking) EDIT: done in #102
- [x] dengue_sensors (need valid locations and names, but API is lacking) EDIT: done in #102
- [x] ecdc_ili EDIT: done in #102
- [x] kcdc_ili (need valid region, but API is lacking) EDIT: done in #102
- [x] paho_dengue (bug prevents testing) EDIT: done in #102

dshemetov:
fixed by #102

PR #56, opened 2022-10-26
change from devtools to remotes for install instructions.

willtownes:
it's better to use remotes since it has fewer dependencies than devtools

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,10 +4,10 @@
 
 ## Install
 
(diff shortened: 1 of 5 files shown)


brookslogan (approved):
Thanks, nice idea!  I applied this change in a few additional (commented) lines here and there.  (+ re-`document()`ed and re-`style_pkg()`ed a few times, which apparently fixed some styling issues that I apparently failed to commit in #54.

PR #74, opened 2023-03-15
Address a few bugs

### LINKED ISSUE #65: Typo: `fip_code`; also, make sure it requires text rather than an integer
`fip_code` should be `fips_code` [here](https://github.com/cmu-delphi/epidatr/blob/46665591b6f37f1e68067db7837f2ef7467657c7/R/endpoints.R#L132).

Additionally, this should require a string rather than an integer / strings rather than integers.  (FIPS codes should include appropriate zero padding, so they should be strings.)

### LINKED ISSUE #66: Output type specification for `covid_hosp_facility` may be wrong
This [line](https://github.com/cmu-delphi/epidatr/blob/e6ccb9a670e0cdde48e662cb82e7b5ec8174d5a9/R/endpoints.R#L178) seems problematic.  Might need to be a `"date"` even though it describes a week.

Test query:
```r
covid_hosp_facility('390119', epirange(20201201, 20201207)) %>% fetch_tbl()
```

### LINKED ISSUE #67: `covid_hosp_facility` does not actually support a list of hospital pks
We should properly validate and document this, rather than returning data for only a single pk.

### LINKED ISSUE #69: Make sure all geo value fields are type `text`

dshemetov:
Closes #69.
Closes #66.
Closes #65.
Closes #67.

On the last issue: request params get formatted differently depending on whether a vector or a list is handed as an argument. This is an attempt to make the outcome uniform. The issue is

```r
r$> paste(sapply(c("290", "190"), toString), collapse = ",")
[1] "290,190"

r$> paste(sapply(list(c("290", "190")), toString), collapse = ",")
[1] "290, 190"
```

So I added some logic to check for vectors and use `as.list` instead. The problem is that the `is.vector` also yields TRUE for lists, so it's currently messing up the list-type handling. Would appreciate help/commits there.


brookslogan:
The function you wanted was probably `is.atomic`.  But I wonder if we can keep the logic a bit simpler, back out your change, and instead fix `toString` to be `paste, collapse=","`.  And maybe also use `vapply` instead of `sapply` (`FUN.VALUE` would be `character(1L)`).

brookslogan (commented):
Looks good, though hope we don't depend on int zips/fips somewhere already.  Committed the approach I described above.  Do you remember what request revealed the `fip_code` typo?  Wonder if we're missing some tests.

brookslogan (approved):

dshemetov:
Thanks Logan! I don't recall what revealed the fip_code typo. To say that we're missing tests is an understatement: those two    functions `format_list` and `format_item` are the only functions that have any tests in this repo at all.

Just going to merge and leave making unit tests to another PR.

Issue #44, opened 2022-09-09
Set up a `gh-pages` branch, host documentation site built using `pkgdown`


brookslogan:
E.g., to match `epiprocess`, `epipredict`, `covidcast`.

There are libraries (maybe `usethis`) to do some/all of this in maybe just a few commands.  @dajmcdon might remember how the two `epi*` packages above were set up.

dajmcdon:
Yeah, the command is `usethis::use_pkgdown_github_pages()`.

dshemetov:
Closed by #78

PR #78, opened 2023-04-01
ci: add docs building and hosting

dshemetov:
Copied example from https://github.com/r-lib/actions/tree/v2/examples#build-pkgdown-site


brookslogan:
Closes #44.  I'll try to review later today.

brookslogan:
For uniformity/convenience, we should try to use `usethis` for all our CI, etc., rather than manually copy-pasting files.  Here, we can use
```r
usethis::use_github_action("pkgdown")
```
This has a slightly different name for the workflow file (committed above).  (It also adds a `.github/.gitignore` ignoring `.html` files (maybe from r-lib/usethis #855) --- I'm not sure I've run into html files being placed there, but maybe it could be from something like nektos/act.)

brookslogan (approved):

PR #79, opened 2023-04-05
feat: add auth functions and add arg to covidcast

dshemetov:
An implementation of API keys:
- [x] API key getter function
- [x] API key setting recipes in README.md
- [x] add auth arg to covidcast call params
- [x] needs some testing


dshemetov:
[Our API supports auth via three approaches](https://delphi-org.slack.com/archives/C02U0AX8KJR/p1681399890622979?thread_ts=1681399526.544069&cid=C02U0AX8KJR): 
1. requests parameter `api_key`,
2. [HTTP Basic authentication](https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication#basic_authentication_scheme): pass a header with "Authentication: Basic [user:pass]" (in our case user is "epidata" and pass is api key)
3. [HTTP Bearer authentication](https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication#authentication_schemes): pass a header with "Authentication: Bearer [pass]" (pass is api key)

I won't implement option 1, so we don't pollute the user's request URLs and reduce chance of them sharing their API key. The latter two seem equivalent, so I'll implement the (3).

R/auth.R
```diff
@@ -0,0 +1,24 @@
+#' @title Get the API key
```
brookslogan, line 1: note: [this page](https://cran.r-project.org/web/packages/roxygen2/vignettes/rd.html) says `@title` is optional unless you are doing something exotic.  The first line of the roxygen docstring should be inferred to be the title.  Same with `@description` below.  (Though `@description` may also be needed if you're adding an `@rdname`/`@describeIn` above it.)

question (non-blocking): Am I missing some benefit of explicit `@title` tags?  (Robust to auto-line-wrapping?)

R/auth.R
```diff
@@ -0,0 +1,24 @@
...
+  warning(
+    paste0(
+      "No API key found. To avoid being rate limited, you can:\n",
+      " - set the environment variable DELPHI_EPIDATA_KEY.\n",
+      " - set the option 'delphi.epidata.key' or\n",
+      "\n",
+      "To save your key for later sessions (and hide it from your code), you can edit your .Renviron file with:\n",
+      "  usewith::edit_r_environ()"
+    )
+  )
```
brookslogan, line 22: suggestion: we could buy into `cli` here to pretty bulleted lists with auto line wrap.  Think it was `cli::cli_bullets`?  I'm not up to speed on all the `cli` features we might consider using; @dajmcdon might have a better idea.  Could be a good topic to go over for a few minutes at a tooling meeting: where / for what we'd want to use `cli` / the most useful functions.

R/request.R
```diff
@@ -9,13 +9,20 @@ join_url <- function(url, endpoint) {
 #' @importFrom httr RETRY
 do_request <- function(url, params) {
   # don't retry in case of certain status codes
-  res <- httr::RETRY("GET", url,
-    query = params, http_headers,
+  res <- httr::RETRY("GET",
+    url,
+    query = params,
+    http_headers,
+    httr::authenticate("epidata", get_auth_key()),
```
brookslogan, line 16: praise: good that the API key isn't a query parameter; I think the special treatment is supposed to be best practice to prevent logging keys all over the place.

question: Should we use `httr::handle`s?  `?httr::authenticate` recommends this, but I don't think its examples actually use it.  Instead, maybe we could mirror the tests [here](https://github.com/r-lib/httr/commit/f0f8de394cc55b66643a48c9f3a156213ea35f86)?  I'm not sure if it's that necessary for us if we re-pass the auth key every time (no chance of it trying to reuse a mistyped key again and again if the user has fixed the env var / option, hopefully).

README.md
```diff
@@ -21,6 +21,18 @@ can be still be installed with the old name by specifying
 
 TODO
 
+## API Keys
+
+The Delphi API will soon require a registered API key for unthrottled queries.
```
brookslogan, line 26: suggestion:
- "registered API key" -> "(free) registered API key" / "(free) API key" --- does "registered" mean associated with someone's background or their name?
- "for unthrottled queries" -> ~"to issue very frequent or complex queries"?  "for unthrottled queries" could read as some special advanced type of query

R/auth.R
```diff
@@ -0,0 +1,24 @@
...
+get_auth_key <- function() {
+  key <- Sys.getenv("DELPHI_EPIDATA_KEY", unset = "")
+  if (key != "") return(key)
+
+  key <- getOption("delphi.epidata.key", default = "")
+  if (key != "") return(key)
+
+  warning(
+    paste0(
+      "No API key found. To avoid being rate limited, you can:\n",
```
brookslogan, line 15: suggestion: mention complex queries here?

R/auth.R
```diff
@@ -0,0 +1,24 @@
...
+#' @return The API key
+#' @export
+get_auth_key <- function() {
+  key <- Sys.getenv("DELPHI_EPIDATA_KEY", unset = "")
+  if (key != "") return(key)
+
+  key <- getOption("delphi.epidata.key", default = "")
+  if (key != "") return(key)
+
+  warning(
```
brookslogan, line 13: suggestion: use `rlang::warn` with `.frequency`, `.frequency_id` options to have a cooldown on the warning.

README.md
```diff
@@ -21,6 +21,18 @@ can be still be installed with the old name by specifying
...
+We recommend storing your key in `.Renviron` file, which R will read by default.
+You can do this with `usethis::edit_r_environ()` and there enter `DELPHI_EPIDATA_KEY=<your-key>`.
+An alternative is [the dotenv package](https://github.com/gaborcsardi/dotenv/).
+
+Note that for the time being, the private endpoints (i.e. those prefixed with `pvt`) will require
+a separate key that needs to be passed as an argument.
+
 ## Development Environment
 
 Relevant R commands
```
brookslogan, line 38: question: were there supposed to be tests in test-auth.R?

R/auth.R
```diff
@@ -0,0 +1,24 @@
...
+    paste0(
+      "No API key found. To avoid being rate limited, you can:\n",
+      " - set the environment variable DELPHI_EPIDATA_KEY.\n",
+      " - set the option 'delphi.epidata.key' or\n",
+      "\n",
+      "To save your key for later sessions (and hide it from your code), you can edit your .Renviron file with:\n",
+      "  usewith::edit_r_environ()"
```
brookslogan, line 20: typo: "usewith" --> "usethis"

README.md
brookslogan: todo: configure CI to use a key.  Else I expect the examples to generate warning messages.  Think we can use GitHub Secrets.  I've played around with those [here](https://github.com/brookslogan/testing-bots-001/blob/main/.github/workflows/label-questions-stale.yml) if you haven't already used them as part of production work.

R/auth.R
brookslogan: praise: really like this error message and interface.

todo: link where to get a key

question: where do I get a testing key?  And can you remind me what the testing server URL is?  (via Slack maybe)

README.md
brookslogan: chore: satisfy the styler CI

R/auth.R
brookslogan: also of interest: using https://cli.r-lib.org/reference/links.html to link directly to page to get a key?

R/auth.R
brookslogan: note: `rlang::warn` has a built-in `cli`-like format which might try to combine & wrap your lines.  (Or maybe that's only if you use actual `cli` via `use_cli_format=TRUE`.)  It might require a little adjustment to properly use its bulleted lists, etc.  But I think it'll enable better-looking messages and have suggested using `cli` below.

brookslogan (changes requested):
Looks nice!  Have a bunch of little questions/suggestions.  Main things:
- add mention of query complexity limit
- typo & tweaks in warning message
- testing and CI stuff

brookslogan:
question: Are the old-epidata-endpoint auth token arguments clearly different from the new api key?  We might need to go through and say something like "an {endpoint name}-specific authentication token (not the same thing as the newer, system-wide API key)".

R/request.R
dshemetov: Well that's frustrating. Why make a function if you're going to recommend working around it?

I tried using handle, but it gave me an error in `parse_url`. I don't really want to spend more time on this, until we actually get a fix request.

```r
  handle <- httr::handle(url)
  res <- httr::RETRY("GET",
    handle = handle,
    query = params,
    terminate_on = c(400, 401, 403, 405, 414, 500),
    http_headers,
    httr::authenticate("epidata", get_auth_key())
  )
```

R/auth.R
dshemetov: I think it was just suggested to me by auto-complete, I'll remove them.

dshemetov:
Yes, they are different. Agreed.

R/auth.R
dshemetov: #86

Issue #43, opened 2022-09-09
Add missing `@export` tags for S3 methods


brookslogan:
See https://github.com/r-lib/devtools/issues/2293.

One important instance (not necessarily the only one) is `print.epidata_call`.

([Also r]equires re-`document`ing and including the [updates --- `NAMESPACE` should be updated to include some S3method entries as a result if the roxygen comments have been appropriately updated].)

PR #81, opened 2023-04-07
docs+dep: another pass and remove fetch_df

### LINKED ISSUE #84: `request_url` handling of default args and other args
`print(call)` did provide "the" `request_url` by default before, but the problem is that the request URL changes based on what further options we provide; see the args to `request_url`.  If you think if a better name for request-URL-with-default-args-for-`request_url`-but-other-arguments-baked-in, please add it!

_Originally posted by @brookslogan in https://github.com/cmu-delphi/epidatr/issues/81#issuecomment-1503716037_

Ideal workflow is something like: `print(call)` by default pretty prints the args and the request url with all the correct arguments set. Implementation details unclear though.

### LINKED ISSUE #85: Typo in flusurv column spec
`rage_age_0` should be `rate_age_0`, etc.  This appears to just cause extra warnings rather than preventing use altogether, so lower priority.
``` r
library(magrittr)
library(epidatr)
flusurv("network_all", epiweeks=epirange(123401, 345601)) %>%
  fetch_tbl()
#> Warning: The following named parsers don't match the column names: rage_age_0,
#> rage_age_1, rage_age_2, rage_age_3, rage_age_4, rage_overall
#> Warning: One or more parsing issues, call `problems()` on your data frame for details,
#> e.g.:
#>   dat <- vroom(...)
#>   problems(dat)
#> # A tibble: 336 × 11
#>    release_date location    issue  epiweek      lag rate_age_0 rate_age_1
#>    <chr>        <chr>       <date> <date>     <int>      <dbl>      <dbl>
#>  1 2017-05-19   network_all NA     2009-08-30   401      0.467      0.316
#>  2 2017-05-19   network_all NA     2009-09-06   400      2.04       0.904
#>  3 2017-05-19   network_all NA     2009-09-13   399      2.04       0.836
#>  4 2017-05-19   network_all NA     2009-09-20   398      2.10       1.31 
#>  5 2017-05-19   network_all NA     2009-09-27   397      4.26       2.01 
#>  6 2017-05-19   network_all NA     2009-10-04   396      7.18       3.70 
#>  7 2017-05-19   network_all NA     2009-10-11   395     10.2        4.38 
#>  8 2017-05-19   network_all NA     2009-10-18   394     10.8        4.09 
#>  9 2017-05-19   network_all NA     2009-10-25   393      8.24       3.59 
#> 10 2017-05-19   network_all NA     2009-11-01   392      5.96       2.30 
#> # ℹ 326 more rows
#> # ℹ 4 more variables: rate_age_2 <dbl>, rate_age_3 <dbl>, rate_age_4 <dbl>,
#> #   rate_overall <dbl>
```

<sup>Created on 2023-04-12 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

dshemetov:
Closes #84, closes #85

* alias epidata_call to create_epidata_call
* link to epidata_call for endpoint returns
* update docs
* \dontrun the private endpoint examples
* removed fetch_df
* print.epidata_call automatically displays the request url for the classic format
* fix typo bug


brookslogan:
Tried to do a simple tweak to the documentation here and it's bringing up check & implementation issues.  We might just want to let the check() S3 notes slide for now, though I'm working to address them.  (I wanted to use `print(call)` in the docs, addressed #43, found that `print(call)` didn't necessarily give the right request URL, `check()` finds that some `print` and `as.data.frame` methods have the wrong signatures, ...)

dshemetov:
Nice changes! `cli` is a nice touch! Question: what do you think of having `print(call)` of providing by default the `request_url`? Or did you try this and run into a bug?

dshemetov:
Also: I really like that you show the piping example in the docs. I didn't want to add magrittr as a dependency, but that's really the workflow we want to model to our users. Maybe magrittr is a light and common enough that, that we can just include it in Suggests and use pipes in all our examples?

brookslogan:
`print(call)` did provide "the" `request_url` by default before, but the problem is that the request URL changes based on what further options we provide; see the args to `request_url`.  If you think if a better name for request-URL-with-default-args-for-`request_url`-but-other-arguments-baked-in, please add it!

Yes, I'd also prefer pipes in the examples.  I did forgot to `Suggests:` magrittr for how I went about this.  Dependency-weight-wise, doesn't matter its weight because we already transitively depend on it via `readr` -> `rlang` -> `tibble` -> `magrittr`.  So maybe we could even just `Imports:` magrittr and re-export the pipe, so the examples don't even need `library(magrittr)`?  In maybe 5 years we might consider the new R built-in pipe instead (`|>`), but I expect some of our potential user base to be stuck on old systems that don't have it available.

dajmcdon:
If you do `usethis::use_pipe()` it will set up the Imports / reexport version for you. That's my usual practice. Until built in pipe becomes common.

dshemetov:
Sounds good, well if we already indirectly depend on it, then I'm happy to just import/export that pipe into the package!

Made a new issue for request_url in #84.

brookslogan (approved):
Somehow my review was stuck as pending.  I think everything I noted + more has been addressed.

PR #89, opened 2023-04-24
Update check functions for clarity and add tests

### LINKED ISSUE #20: pvt_paho_dengue points to the wrong endpoint
`paho_dengue` points to the quidel endpoint.

### LINKED ISSUE #36: Fix `issue` parsing in `fluview` endpoint
It looks like `issue` is set to be parsed as a date when it's actually an epiweek; the result is `NA` issues:

``` r
  library(delphi.epidata)
  library(magrittr)
  fluview("hhs1", epirange(201501,201503)) %>% fetch_tbl()
#> Warning: One or more parsing issues, see `problems()` for details
#> # A tibble: 3 × 16
#>   release_date region issue  epiweek      lag num_ili num_patients
#>   <chr>        <chr>  <date> <date>     <int>   <int>        <int>
#> 1 2017-10-24   hhs1   NA     2015-01-04   143     915        48019
#> 2 2017-10-24   hhs1   NA     2015-01-11   142    1306        51021
#> 3 2017-10-24   hhs1   NA     2015-01-18   141    1923        50890
#> # … with 9 more variables: num_providers <dbl>, num_age_0 <int>,
#> #   num_age_1 <int>, num_age_2 <int>, num_age_3 <int>, num_age_4 <int>,
#> #   num_age_5 <int>, wili <dbl>, ili <dbl>
  print(readr::problems())
#> # A tibble: 3 × 5
#>     row   col expected         actual file                            
#>   <int> <int> <chr>            <chr>  <chr>                           
#> 1     2     3 date like %Y%m%d 201740 /tmp/Rtmp2ckL6r/file40877e0d5bbf
#> 2     3     3 date like %Y%m%d 201740 /tmp/Rtmp2ckL6r/file40877e0d5bbf
#> 3     4     3 date like %Y%m%d 201740 /tmp/Rtmp2ckL6r/file40877e0d5bbf
```

<sup>Created on 2022-07-13 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1) --- and fixed up due to `readr::problems()` somehow being an empty 0x4 tibble without the `file` column </sup>

### LINKED ISSUE #46: Fix error in epirange list check
- [ ] Fix [this line](https://github.com/cmu-delphi/epidatr/blob/68dcae4609d170ae3df9a6a0319065f211c5be53/R/check.R#L57) to call `is.list` on the appropriate argument and change the non-working nested `sapply`s to something that makes sense (probably a single `purrr::map_lgl`).  I encountered this when trying to feed a single `Date` object in.
- [ ] Fix any downstream behavior, e.g., in preparing the query, which may not actually accommodate, e.g., a list of epiranges.
- [ ] Add tests for this case.
- [ ] Update docs.

### LINKED ISSUE #48: `covidcast`'s `as_of` should not "require" an epirange, and does not properly handle epiranges
See line [here](https://github.com/cmu-delphi/epidatr/blob/68dcae4609d170ae3df9a6a0319065f211c5be53/R/endpoints.R#L690).  `as_of` in the covidcast package allows only a single date; ~we should support that in epidatr~ [it looks like we already support this in epidatr, but we should make sure to test we can use a `Date` class here after #47 is fixed, and/or clarify the error message that an "epirange" is required].  Further, epiranges aren't actually handled properly here; providing a non-zero-width epirange only produces results for one `as_of` (plus no column for which `as_of` it came from --- note that `as_of` and `issue` are not interchangeable).  Maybe actually handling epiranges could be made into a separate "enhancement" request issue; for now, maybe we should just reject epiranges and allow only single dates for `as_of`.

<del>
Currently working around for the single-as-of case like this:

```r
  epidatr::covidcast("hhs", "confirmed_admissions_influenza_1d", "day", geo_type,
                     epidatr::epirange(format(target_date_range[[1L]], "%Y%m%d"),
                                       format(target_date_range[[2L]], "%Y%m%d")),
                     geo_values,
                     as_of=epidatr::epirange(format(evaluation_as_of, "%Y%m%d"),
                                             format(evaluation_as_of, "%Y%m%d"))
                     ) %>%
    epidatr::fetch_tbl()
```
</del>

``` r
library(epidatr)
library(magrittr)
generates_misleading_error =
  epidatr::covidcast("hhs", "confirmed_admissions_influenza_1d",
                     "day", "nation",
                     epidatr::epirange("20220801", "20220809"),
                     as_of = as.Date("2022-09-09"),
                     "us") %>%
  fetch_tbl()
#> Error in `check_single_epirange_param()`:
#> ! argument as_of is not a epirange
works_but_could_be_shorter =
  epidatr::covidcast("hhs", "confirmed_admissions_influenza_1d",
                     "day", "nation",
                     epidatr::epirange("20220801", "20220809"),
                     as_of = format(as.Date("2022-09-09"), "%Y%m%d"),
                     "us") %>%
  fetch_tbl()
wrong =
  epidatr::covidcast("hhs", "confirmed_admissions_influenza_1d",
                     "day", "nation",
                     epidatr::epirange("20220801", "20220809"),
                     as_of = epidatr::epirange(
                       format(as.Date("2022-09-09"), "%Y%m%d"),
                       format(as.Date("2022-09-20"), "%Y%m%d")
                     ),
                     "us") %>%
  fetch_tbl()
all.equal(works_but_could_be_shorter, wrong) # TRUE is bad
#> [1] TRUE
```

<sup>Created on 2022-09-27 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### LINKED ISSUE #62: `covid_hosp_facility` example yields empty string
I am having trouble constructing a call that doesn't give an empty string as a result.  We should fix this example / the implementation, and also move examples to use `fetch_tbl` instead of `fetch_csv` so this situation will raise an error (plus, as mentioned independently, to be more relevant to users).

### LINKED ISSUE #68: `epirange` interface does not support `time_value=*`

### LINKED ISSUE #94: Make a "timeset" help topic
question: do we have a "timeset" help topic to link to instead?

_Originally posted by @brookslogan in https://github.com/cmu-delphi/epidatr/pull/26#discussion_r1181951867_

dshemetov:
This PR rewrites all the `check_*` functions with `checkmate` and adds tests.

Changes:
- consolidate `check_single_int_param, check_int_param`-> `check_int_like_param`
- consolidate `check_single_string_param, check_string_param` -> `check_character_param`
- add `check_date_param`
- consolidate `check_single_epirange_param, check_epirange_param` -> `check_timeset_param`
- update `check_timeset_param` to correctly detect a list of epiranges and to correctly error when given a vector of epiranges

Fixes:
- fixes #48
- fixes #46
- fixes #36
- fixes #62 (I think this was fixed already, but tagging this in here so we remember to close the issue)
- fixes #68
- fixes #20 
- fixes #94


dajmcdon:
I'm not going to block, but it's probably worth examining the related implementation of similar checks in [`{epipredict}`](https://github.com/cmu-delphi/epipredict/blob/9a804303d97aa9d3de1323c66a52797acce86483/R/utils-arg.R).

@lcbrooks it may be worth putting these kind of potentially shared functions somewhere common.

dshemetov:
I took a glance at `checkmate` and decided not to use it here for the primary reason that it doesn't remove hard decisions about what data types you allow in your functions. So I focused on making those decisions instead of learning a new package.

`checkmate` might still help in `epipredict`, `epiprocess`, or even here, in a later refactor.

brookslogan:
Still need to give this a look over.  Brief partial skim: is #36 actually addressed?  Or is the code sample showing the old behavior?

dshemetov:
Ah good catch, I misunderstood that issue. It's not fixed yet.

R/check.R
```diff
@@ -1,90 +1,186 @@
...
 }
 
-check_single_epirange_param <- function(name, value, required = TRUE) {
-  if ((required &&
-    !(!is.null(value) &&
-      is_epirange_like(value))) ||
-    (!required && !(is.null(value) || is_epirange_like(value)))) {
-    rlang::abort(paste0("argument ", name, " is not a epirange"),
+#' Check if param is datelike
+check_single_date_param <- function(name, value, required = TRUE) {
```
brookslogan, line 101: suggestion: `date` -> `date_like`

R/check.R
```diff
@@ -1,90 +1,186 @@
...
+  if (is.null(value)) {
+    if (required) {
+      rlang::abort(paste0("argument ", name, " is not a datelike"),
```
brookslogan, line 104: suggestion: "datelike" -> "date_like", in hopes that `??date_like` will do something useful

R/check.R
```diff
@@ -1,90 +1,186 @@
...
   }
 }
 
-check_epirange_param <- function(name, value, required = TRUE) {
+#' Check if param is datelike or vector of datelike
+check_date_param <- function(name, value, required = TRUE) {
   if (is.null(value)) {
     if (required) {
-      rlang::abort(paste0("argument ", name, " is not a epirange"),
+      rlang::abort(paste0("argument ", name, " is not a datelike"),
```
brookslogan, line 121: suggestion: on all these NULL cases, say something like "argument {name} should be a {blah}, but was NULL/missing".  Currently they sort of suggest that the user explicitly fed in something of the wrong type, but I'm assuming we're just getting NULLs as default args.

R/check.R
```diff
@@ -1,90 +1,186 @@
...
+#' Return TRUE if value is a character or numeric or EpiRange
+#' TODO: add Date support
+is_epirange_like <- function(value) {
+  is.character(value) ||
+    is.numeric(value) ||
+    inherits(value, "EpiRange")
+}
+
+#' Check if param is a character or numeric or EpiRange
+check_single_epirange_param <- function(name, value, required = TRUE) {
```
brookslogan, line 147: suggestion: `epirange` -> `epirange_like`

R/check.R
```diff
@@ -1,90 +1,186 @@
...
+      )
+    }
+    return()
+  } else if (is_epirange_like(value)) {
+    return()
+  } else if (all(vapply(value, is.list, logical(1L))) && length(value) > 0) {
+    for (i in seq_along(value)) {
+      check_single_epirange_param(paste0(name, "[", i, "]"), value[[i]], required = TRUE)
+    }
+  } else if (sum(names(value) %in% c("from", "to")) >= 2) {
```
brookslogan, line 178: polish: eliminate the from&to part here since it was removed from the single epirange-like above.

R/check.R
```diff
@@ -1,90 +1,186 @@
...
-check_single_int_param <- function(name, value, required = TRUE) {
-  if ((required &&
-    !(!is.null(value) &&
-      is.numeric(value) &&
-      length(value) == 1)) ||
-    (!required &&
-      !(is.null(value) || (is.numeric(value) && length(value) == 1)))) {
-    rlang::abort(paste0("argument ", name, " is not a single integer"),
+#' Check if param is a character or numeric or EpiRange
+check_epirange_param <- function(name, value, required = TRUE) {
```
brookslogan, line 164: suggestion: use a different naming pattern from what we use for actual vectorized structures.  E.g., `check_single_or_list_epirange_like_param`, `check_time_set_param`, `check_time_spec_param`, etc.

tests/testthat/test-check.R
```diff
@@ -0,0 +1,132 @@
+test_that("check_single_string_param", {
```
brookslogan, line 1: praise: great to have all these tests, covering both error & nonerror behavior!  And to see date_likes and epirange_likes distinguished.

R/check.R
```diff
@@ -1,90 +1,186 @@
...
 
+#' Allows a single numeric value
+check_single_int_param <- function(name, value, required = TRUE) {
+  if (is.null(value)) {
+    if (required) {
+      rlang::abort(paste0("argument ", name, " is not a single number"),
+        name = name, value = value, class = "invalid_argument"
+      )
+    }
+  } else if (is.numeric(value)) {
```
brookslogan, line 55: issue: this isn't a check that it's an integer.  If we already use rlang then I'd suggest `rlang::is_integerish` and also call this check `integerish`.  If we really mean integer-values-specifically-with-integer-type, then `is.integer` rather than `is.numeric`.  + update "number" below to describe whatever notion of integer is settled on.
OR: rename `int` checks to, e.g., `numeric`.  If you want to use `checkmate`'s terminology, a single numeric would be a `number`.

brookslogan (commented):
I have a lot of naming questions and polish comments which could be debated/deferred, but already this is a naming improvement and takes care of a few bugs.  I think the current logic and naming doesn't match quite well with R's always-vectors approach, naming conventions (helped a lot by your addition of these "single" checker functions), and check function conventions (I think missingness would be handled separately).

thought: I believe [some] related rlang functions combine handle the single/"scalar" case simply by adding on an `n` argument to the vector case.  Though `checkmate` takes a different approach and tries to use different names and functions for [additional] single/scalar checks.  If you feel it'd be more maintainable to reduce the number of functions, you might try the rlang approach.

question: I haven't really given a good check over any actual functionality issues or the tests, but the tests seem pretty extensive so they might already have caught everything.  Are you confident this behaves well or should I give it a second pass trying to find bugs?

R/check.R
```diff
@@ -1,90 +1,186 @@
...
+    }
+  } else {
+    rlang::abort(paste0("argument ", name, " is not a single string"),
+      name = name, value = value, class = "invalid_argument"
+    )
+  }
+}
+
+#' Allows string vectors of length > 1 and lists of string vectors of length 1
 check_string_param <- function(name, value, required = TRUE) {
```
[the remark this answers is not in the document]
brookslogan: rlang uses "string" to describe a single string, "character" for character vectors of any (or, optionally one specific) length

dajmcdon:
On naming: perhaps "scalar" rather than "single"?

R/check.R
```diff
@@ -1,90 +1,186 @@
...
-    (!required && !(is.null(value) || is.character(value)))) {
+  if (is.null(value)) {
+    if (required) {
+      rlang::abort(paste0("argument ", name, " is not a string"),
+        name = name, value = value, class = "invalid_argument"
+      )
+    }
+  } else if (is.character(value)) {
+    return()
+  } else if (is.list(value)) {
```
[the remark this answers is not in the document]
dshemetov: I added the list option for consistency with epirange, where we have to pass lists. I'm down for whatever is most natural for R users though.

R/check.R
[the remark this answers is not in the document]
dshemetov: I think the goal of this function is to create a field that allows both of these examples: `geo_values=c("ak", "ca")` and `geo_values="ak,ca"`.

R/check.R
dshemetov: I changed these checks to `is.integer` and found that it returned FALSE for `value = 1`, which I could see being annoying. So I stuck with `is.numeric`, but I agree, it's too flexible.

Integerish seems perfect for this, so I'll change to that, thanks.

R/check.R
```diff
@@ -1,90 +1,186 @@
...
-  if ((required &&
-    !(!is.null(value) &&
-      is.numeric(value))) ||
-    (!required && !(is.null(value) || is.numeric(value)))) {
+  if (is.null(value)) {
+    if (required) {
+      rlang::abort(paste0("argument ", name, " is not a number"),
+        name = name, value = value, class = "invalid_argument"
+      )
+    }
```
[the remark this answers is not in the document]
dshemetov: Sounds good

R/check.R
dshemetov: I like `time_set_param`

R/check.R
dshemetov: This is my round-about way of detecting when someone gives us a vector of epiranges, which becomes a mangled list:
```
r$> c(epirange(1, 2), epirange(3, 4))
$from
[1] 1

$to
[1] 2

$from
[1] 3

$to
[1] 4
```
Is there a better way to handle that?

dshemetov (commented):
Thanks Logan and Dan! Going to 
- make some conversions to `checkmate`
- rename `single` -> `scalar`
- add `Date` support

R/endpoints.R
```diff
@@ -701,15 +701,15 @@ covidcast <-
...
-    check_single_string_param("geo_type", geo_type)
-    check_epirange_param("time_values", time_values)
-    check_string_param("geo_values", geo_values)
-    check_single_epirange_param("as_of", as_of, FALSE)
-    check_epirange_param("issues", issues, FALSE)
-    check_single_int_param("lag", lag, FALSE)
+    check_character_param("data_source", data_source)
+    check_character_param("signals", signals)
+    check_character_param("time_type", time_type)
+    check_character_param("geo_type", geo_type)
```
brookslogan, line 707: question: original "single string" and "string" checks are replaced by "character" check here.  Is this okay, or should there be a (rlang-definition) string/scalar_character check?

R/check.R
```diff
@@ -1,91 +1,80 @@
...
-      length(value) == 1)) ||
-    (!required &&
-      !(is.null(value) || (is.numeric(value) && length(value) == 1)))) {
-    rlang::abort(paste0("argument ", name, " is not a single integer"),
-      name = name, value = value, class = "invalid_argument"
-    )
-  }
+#' Allows a scalar timeset param: a character, an integerish, or an EpiRange
+#' @importFrom checkmate assert check_character check_integerish check_class
+check_scalar_timeset_param <- function(name, value, required = TRUE) {
```
brookslogan, line 57: Can we call these `assert_*` or whatever the checkmate equivalent is?  checkmate uses `check_*` to indicate the output is TRUE/FALSE.

R/endpoints.R
brookslogan: (same question for other endpoints as well..)

R/check.R
```diff
@@ -1,91 +1,80 @@
...
-    )
-  }
+#' Allows a scalar date_like param: a single date, character, or integerish
+#' @importFrom checkmate assert check_date check_character check_integerish
+check_scalar_date_param <- function(name, value, required = TRUE) {
+  null.ok <- !required
+  assert(
+    # check_date(value, null.ok = null.ok, len = 1),
+    check_character(value, null.ok = null.ok, len = 1),
+    check_integerish(value, null.ok = null.ok, len = 1),
```
brookslogan, line 36: todo: Date support should allow `Date` objects here

R/check.R
```diff
@@ -1,91 +1,80 @@
...
-    )
-  }
+#' Allows a vector of date_like params: dates, characters, or integerishes
+#' @importFrom checkmate check_date check_character check_integerish
+check_date_param <- function(name, value, required = TRUE) {
+  null.ok <- !required
+  assert(
+    # check_date(value, null.ok = null.ok),
+    check_character(value, null.ok = null.ok),
+    check_integerish(value, null.ok = null.ok),
```
brookslogan, line 49: todo: ditto here

tests/testthat/test-check.R
```diff
@@ -0,0 +1,109 @@
...
+  # expect_no_error(check_single_date_param("name", as.Date("2020-01-01")))
+  # expect_no_error(check_single_date_param("name", as.Date("2020-01-01"), required = FALSE))
+  expect_error(check_scalar_date_param("name", NULL))
+  expect_no_error(check_scalar_date_param("name", NULL, required = FALSE))
+  expect_error(check_scalar_date_param("name", c(20200101, 20200101)))
+  expect_error(check_scalar_date_param("name", list(20200101, 20200101)))
+  # expect_error(check_single_date_param("name", c("2020-01-01", 20200102, as.Date("2020-01-03"))), "argument is not a datelike")
+  # expect_error(check_single_date_param("name", c("2020-01-01", 20200102, as.Date("2020-01-03")), required = FALSE), "argument is not a datelike")
+  # expect_error(check_single_date_param("name", list("2020-01-01", 20200102, as.Date("2020-01-03"))), "argument is not a datelike")
+  # expect_error(check_single_date_param("name", list("2020-01-01", 20200102, as.Date("2020-01-03")), required = FALSE), "argument is not a datelike")
```
brookslogan, line 61: chore: clean up commented-out stuff once the new interface is all set

tests/testthat/test-check.R
```diff
@@ -0,0 +1,109 @@
+test_that("check_string_param", {
+  expect_no_error(check_string_param("name", "value"))
+  expect_no_error(check_string_param("name", "value", required = FALSE))
```
brookslogan, line 3: praise: I like the way this reads with the `required = FALSE` spelled out.  (And also like the `expect_no_error` over the old arcane way of `expect_error({expr}, regexp = NA)` or whatever it was.)

R/check.R
brookslogan: todo: test endpoint/request behavior with hyphenated date strings and Dates.  I expect this to break.  Fix for Dates is probably `nonhyphenated_datestring <- format(date, "%Y%m%d")`.  And hyphenated strings could either convert to Date and do the same, or substring to get the nonhyphenated form (the former would be stricter client-side, as conversion to Date would do some extra validation).

brookslogan (changes requested):
Awesome to see these simplifications!  A couple remaining important things:
* Make sure we check for single strings / scalar characters appropriately.  (Note: I think I've been unclear here... rlang's "string" is a length-1 non-NA character vector, not just a length-1 character vector.  We probably want to be forbidding NAs in just about everything, but not a mandatory thing to address right now.)
* Make sure Date support actually works in requests.

R/endpoints.R
dshemetov: Yea, I made a mistake here, fix incoming!

tests/testthat/test-check.R
```diff
@@ -0,0 +1,109 @@
...
+  expect_no_error(check_scalar_timeset_param("name", 20200101))
+  expect_no_error(check_scalar_timeset_param("name", 20200101, required = FALSE))
+  expect_no_error(check_scalar_timeset_param("name", epirange(20200101, 20200102)))
+  expect_no_error(check_scalar_timeset_param("name", epirange(20200101, 20200102), required = FALSE))
+  # expect_no_error(check_single_epirange_param("name", as.Date("2020-01-01")))
+  # expect_no_error(check_single_epirange_param("name", as.Date("2020-01-01"), required = FALSE))
+  expect_error(check_scalar_timeset_param("name", NULL))
+  expect_no_error(check_scalar_timeset_param("name", NULL, required = FALSE))
+  expect_error(check_scalar_timeset_param("name", c("2020-01-01", 20200102)))
+  expect_error(check_scalar_timeset_param("name", list("2020-01-01")))
```
brookslogan, line 90: suggestion: add tests for "*"

brookslogan:
Awesome, think it's 99% there.

issue: hyphenated date strings don't work.  (Or are they supposed to?)
```{r}
  uiop = covidcast(
     data_source = "jhu-csse",
     signals = "confirmed_7dav_incidence_prop",
     time_type = "day",
     geo_type = "state",
     time_values = epirange("2020-06-01", "2020-08-01"),
     geo_values = "ca,fl"
   ) %>% fetch_tbl()
```

A few things that could be refiled into Issues / other PRs:

todo (non-blocking): add yourself as an author/contributor in DESCRIPTION & redoc.  And Kean Ming as well if he's not already there.
issue (non-blocking): the no-API-key warning
```
To avoid this, you can get your key [here] and then:
ℹ set the environment variable DELPHI_EPIDATA_KEY
ℹ set the option 'delphi.epidata.key'
```
should have something like
```
DELPHI_EPIDATA_KEY, or
```
or
```
DELPHI_EPIDATA_KEY, and/or
```
to indicate you don't have to do both.

dajmcdon:
To follow on Logan's previous comments:

1. I would hope that hyphenated date strings _would_ work with `epirange()` (but fine to add an issue for later).
2. I would update the message to say exactly how to set the API key. Something like `Call Sys.setenv(DELPHI_EPIDATA_KEY = blah)`

dshemetov:
I can probably make hyphenated date strings work with epirange, but there's currently a [server bug that makes them not work](https://github.com/cmu-delphi/delphi-epidata/issues/1153).

dajmcdon:
Can they just be processed to numeric on the R side?

dshemetov:
So far I have avoided processing character / integer inputs as dates, hoping that we can rely on the API to return validation errors. But I could give date parsing a shot.

dshemetov:
Added a date validation function. This query works now

```r

r$> modified_example = covidcast( 
           data_source = "jhu-csse", 
           signals = "confirmed_7dav_incidence_prop", 
           time_type = "day", 
           geo_type = "state", 
           time_values = list(epirange("2020-06-01", "2020-06-03"), epirange("2020-07-01", "20200703")),
           geo_values = "ca,fl" 
         ) %>% fetch_tbl() 
                                                                                                                                                                                                                                                  
r$> modified_example
# A tibble: 12 × 15
   geo_value signal                        source   geo_type time_type time_value direction issue        lag missing_value missing_stderr missing_sample_size value stderr sample_size
   <chr>     <chr>                         <chr>    <ord>    <ord>     <date>         <dbl> <date>     <int>         <int>          <int>               <int> <dbl>  <dbl>       <dbl>
 1 ca        confirmed_7dav_incidence_prop jhu-csse state    day       2020-06-01        NA 2023-03-10  1012             0              5                   5  6.84     NA          NA
 2 fl        confirmed_7dav_incidence_prop jhu-csse state    day       2020-06-01        NA 2023-03-03  1005             0              5                   5  3.34     NA          NA
 3 ca        confirmed_7dav_incidence_prop jhu-csse state    day       2020-06-02        NA 2023-03-10  1011             0              5                   5  6.83     NA          NA
 4 fl        confirmed_7dav_incidence_prop jhu-csse state    day       2020-06-02        NA 2023-03-03  1004             0              5                   5  3.41     NA          NA
 5 ca        confirmed_7dav_incidence_prop jhu-csse state    day       2020-06-03        NA 2023-03-10  1010             0              5                   5  6.66     NA          NA
 6 fl        confirmed_7dav_incidence_prop jhu-csse state    day       2020-06-03        NA 2023-03-03  1003             0              5                   5  4.03     NA          NA
 7 ca        confirmed_7dav_incidence_prop jhu-csse state    day       2020-07-01        NA 2023-03-10   982             0              5                   5 17.1      NA          NA
 8 fl        confirmed_7dav_incidence_prop jhu-csse state    day       2020-07-01        NA 2023-03-03   975             0              5                   5 32.9      NA          NA
 9 ca        confirmed_7dav_incidence_prop jhu-csse state    day       2020-07-02        NA 2023-03-10   981             0              5                   5 18.0      NA          NA
10 fl        confirmed_7dav_incidence_prop jhu-csse state    day       2020-07-02        NA 2023-03-03   974             0              5                   5 36.2      NA          NA
11 ca        confirmed_7dav_incidence_prop jhu-csse state    day       2020-07-03        NA 2023-03-10   980             0              5                   5 17.6      NA          NA
12 fl        confirmed_7dav_incidence_prop jhu-csse state    day       2020-07-03        NA 2023-03-03   973             0              5                   5 36.6      NA          NA
```

R/epidatacall.R
```diff
@@ -86,9 +87,9 @@ request_arguments <-
     for (name in names(all_params)) {
       v <- all_params[[name]]
       if (!is.null(v)) {
-        if (inherits(v, "EpiRange")) {
+        if (test_class(v, "EpiRange")) {
           formatted_params[[name]] <- format_item(v)
-        } else if (is.list(v)) {
+        } else if (test_list(v) || test_vector(v)) {
```
brookslogan, line 92: 1. This looks a little weird because lists are already vectors.
2. Also, "scalars" are vectors, so I don't see us getting to the else branch below.

I think point 2. might save us from `format_item` not working correctly on a length>1 Date vector, but from the last branch in `format_item` is looks like it was intended to work on vectors.  (But we would probably break on the weird usage `list(my_epirange, my_additional_date_vector)`

`format_list` and `format_item` are a little unclear in this vectorized world.

I'll try to update in light of above.

R/epidatacall.R
```diff
@@ -86,9 +87,9 @@ request_arguments <-
     for (name in names(all_params)) {
       v <- all_params[[name]]
       if (!is.null(v)) {
-        if (inherits(v, "EpiRange")) {
+        if (test_class(v, "EpiRange")) {
```
brookslogan, line 90: note: probably doesn't matter here, because we shouldn't be running this in a tight loop, but `test_*` functions look very slow, e.g., 4x--17x the time.  Due to (a) being implemented on top of check functions that also might prepare an error message (which `test_*` converts to `FALSE`), and (b) that have more features (that we don't use, but which probably have significant cost relative to the basic check even when not used).

dshemetov: Right, I am choosing consistency here

tests/testthat/test-check.R
```diff
@@ -0,0 +1,46 @@
...
+})
+
+test_that("assert_timeset_param", {
+  expect_no_error(assert_timeset_param("name", "2020-01-01", len = 1))
+  expect_error(assert_timeset_param("name", c("2020-01-01", "2021-01-02"), len = 1))
+  expect_no_error(assert_timeset_param("name", c("2020-01-01", "2021-01-02")))
+  expect_no_error(assert_timeset_param("name", c(20200101, 20200101)))
+  expect_no_error(assert_timeset_param("name", c(as.Date("2020-01-01"), as.Date("2020-01-02"))))
+  expect_no_error(assert_timeset_param("name", epirange(20200101, 20200102)))
+  expect_error(assert_timeset_param("name", c(epirange(20200101, 20200102), epirange(20200103, 20200104))))
```
brookslogan, line 41: issue: we don't want to allow this.  Right now our EpiRange class doesn't react well inside `c`, and I believe this would actually end up asking for 4 dates individually, not 2 ranges.  The list-of-epiranges is currently the way to go.  We might be able to make things more robust with a more featureful EpiRange class, but it's seems like a lot of work for something that seems like an edge case.  Might prefer just banning lists altogether (but keeping Date vectors and single EpiRanges as options).

brookslogan:
Pushed some changes that I believe resolve the remaining comments.  Please check over & merge.

brookslogan (approved):
I can't seem to request @dshemetov to review my tweaks since he's the PR author.  Just marking approved.

tests/testthat/test-check.R
dshemetov: #97

PR #99, opened 2023-05-06
Consolidate `fetch` interfaces

### LINKED ISSUE #17: Ambiguous error when using fetch_tbl on a null response

### LINKED ISSUE #58: API call returns data frame instead of string
Currently calling the fetch_csv function returns a string representation of the CSV, it would make it easier for users if it was automatically parsed into a data frame or some other appropriate data structure.

### LINKED ISSUE #72: Consolidate `fetch_*` interfaces
- Combine `fetch_df, fetch_classic, fetch_tbl` into a single `tibble` fetcher
- Print a suggestion to use one of the above when printing an EpidataCall object

### LINKED ISSUE #83: Some fetching functions don't appropriately stop/warn on epidata msgs/status
We stop for httr status codes, but not Epidata errors / status codes.  For example, we don't stop for the "no results" error.  Testing out a few fetching functions:
``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
library(epidatr)
fluview("bogus", epirange(201501,201503)) %>%
  fetch_classic()
#> $epidata
#> data frame with 0 columns and 0 rows
#> 
#> $result
#> [1] -2
#> 
#> $message
#> [1] "no results"
fluview("bogus", epirange(201501,201503)) %>%
  fetch_json()
#> data frame with 0 columns and 0 rows
fluview("bogus", epirange(201501,201503)) %>%
  fetch_tbl()
#> Error: '' does not exist in current working directory ('/tmp/RtmpWbkEDL/reprex-1ba5e781fc6ae-ripe-human').
fluview("bogus", epirange(201501,201503)) %>%
  fetch_csv()
#> [1] ""
#> attr(,"class")
#> [1] "epidata_csv" "character"
```

<sup>Created on 2023-04-10 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

So `fetch_json` has this problem, and probably `fetch_tbl` and `fetch_csv` too, though it may not be clear because of #17.  (`fetch_classic` is fine, as it's representing the error directly and not throwing it away, though I'm not sure we want that 0-row 0-column data frame or something else.)

### LINKED ISSUE #88: classic only endpoints should return plain JSON, but fromJSON auto converts to data.frame
Bug: `delphi` *should* be a list / nested structure, but `fetch_classic` doesn't actually respect this and turns it into a data frame.  The culprit is probably
```{r}
  r <- httr::content(res, "text", encoding = "UTF-8")
  m <- jsonlite::fromJSON(r)
```
vs. delphi_epidata.R's
```{r}
content(res, 'parsed')
```

_Originally posted by @brookslogan in https://github.com/cmu-delphi/epidatr/issues/72#issuecomment-1515568257_

### LINKED ISSUE #106: `fetch_tbl()` warning from `vroom` sometimes (when `NA`'s are present?)
```r
covid_hosp_state_timeseries(states = "MA", dates = "20200510") %>% fetch_tbl()

## Warning: One or more parsing issues, call `problems()` on your data frame for details,
## e.g.:
##   dat <- vroom(...)
##   problems(dat)
```

There are a number of calls where this happens, it appears to be `NA`'s, not exactly sure. Will document more examples soon

```r
cm <- covidcast_meta() %>%
  fetch_tbl()
```

```r
fluview_meta() %>% fetch_tbl()
```

dshemetov:
A list of changes:
* add the `fetch` function, which most of the time calls `fetch_tbl` and calls `fetch_classic` if the endpoint is `only_supports_classic` (this is `delphi`, `pvt_norostat_meta`, and `meta`)
* `fetch_tbl` now uses `fetch_classic` to make requests (instead of CSV)
* unexport `fetch_csv`, `fetch_classic`, `fetch_tbl`
* remove `fetch_json`
* add unexported `fetch_debug` for response debugging
* `request_impl` now always raises on HTTP errors, forwards the HTML body to the error
* `fetch` always raises for epidata errors, including no results
* test and check for differences in the tibbles output by `fetch_tbl` and `fetch_csv` (identical on a few small queries)
* many updates to documentation and vignettes
* add mockery and mockr to Suggests
* add new import xml2 to Depends
* export as_tibble

TODO:
- [ ] add more a few more unit tests to the `fetch` interfaces

Fixes:
- fixes #72 
- fixes #88 
- fixes #83
- fixes #58 
- fixes #17 
- fixes #106 (not sure how, might be related to switching away from readr::read_csv to jsonlite::fromJSON)


R/model.R
```diff
@@ -95,17 +95,21 @@ create_epidata_field_info <- function(name,
...
-  }
-  if (info$type == "bool") {
+  } else if (info$type == "bool") {
     return(as.logical(value))
+  } else if (info$type == "int") {
+    return(as.integer(value))
+  } else if (info$type == "float") {
+    return(as.double(value))
+  } else if (info$type == "categorical") {
+    return(as.ordered(value))
```
brookslogan, line 112: question: should this be `as.factor`?  or do our categorical things typically have a natural order to them?

brookslogan (changes requested):
praise: interface is looking more streamlines and slimmer
praise: looks like people can't miss epidata errors now!
issue: need to warn on epidata warnings in message.
todo: figure out whether epidata statuses or http statuses should be checked first.  (see API keys doc for key situation.  another one is when results are truncated, especially in older endpoints where the limits may be very small; maybe this should actually be an error because there's not a clear bug-free way to take advantage of any of the results, unless someone just wanted a preview, but then they could/should just ask for less data)
question: what is return code 2?  is that for truncated results?
todo: fix "1 for success, non-zero for failure"
question: (answer depends on some of the above questions/changes) what's the idea for `fetch_classic()` regarding errors and warnings?  Right now it seems like a kind of mixed thing that stops in some cases, but still returns status&message info when it succeeds.  Can we make it more consistent?  (E.g., try to stop on all errors, warn on all warnings, and return `$epidata`?)
suggestion: rename "`method = c("data.frame", "csv")` to something like "`via = c("classic", "csv")` or `encoding = c("classic", "csv")`
question: is "fixtures" a standard term for these test prototype objects?  `testthat` docs seem to use "fixture" to refer to `local_*()` functions
issue: (potentially; might be server thing) `delphi(system = "ec", epiweek = 202006) %>% fetch()` and `delphi(system = "ec", epiweek = 202006) %>% fetch_classic()` return 500s
question: Sorry, I'm so confused right now.  I see in test comments and now code that "fetch_classic uses jsonlite::fromJSON, which converts the underlying data to a data.frame".  But I thought we wanted this to give a nested list format on the only-classic things; we'd just want the old client's way of getting content for this function, and only use fromJSON when `fetch_tbl`-ing via the `"classic"` transfer-format/encoding, right?  Or am I misunderstanding what `fetch_classic` output-format is supposed to be now?
todo: clarify/fix "fetch_csv uses readr::read_csv"; I assume it's `fetch_tbl(...., format = "csv")` that does this

Stopping here because I'm confused and may just be writing unhelpful things at this point.  I haven't given the test based on the fixture objects a good look through yet, think I'll just muddle things further.

note: okay, now I see the `disable_data_frame_parsing` parameter.  But I assume some statements I mentioned above may need clarified by mentioning this parameter or something along those lines.
note: I also committed a few minor changes to do with a dplyr deprecation and what I think current testthat practices are, roxygen tweaks, various wording tweaks

dshemetov:
- wrt epidata warnings: do you know what status codes epidata warnings have? I'm currently only checking for status code 1 or -2

R/model.R
dshemetov: yes, so i used `as.factor` before, but that didn't match with our CSV parser, which returned both factor and ordered types, so I tried this.

but now this made me realize a better solution here:
`factor(value, levels = info$categories, ordered = TRUE)`

which returns identical output to our CSV parser

brookslogan:
A couple error/warning cases I can think of:
- Truncated results due to row limit.  This should at least have a non-"success" message, but I don't know if there's a standard result code.
- I see a line `if (r$result != 1 && r$result != 2) {` so I assume something's returning 2 as a non-error thing.  Maybe this is the result code for the truncated results?
- Probably any non-"success" message should be turned into an R message or warning.
- I'm not sure what will happen with API keys.  We say we will return something like a 401 or 429 but also say there will be an epidata message.  I believe both are simultaneously possible (returning an http error code + also content).  But if httr stop_for_status stops us early, we might not show the user the explanatory error message.
- Don't know of any codes between 200 and 400 exclusive, but that's just because I have no memory.

brookslogan:
Regarding fixtures: Thanks for the link.  I had read some pytest fixture description before but it just sounded like testthat's definition, but now looking at examples, they do look more like fixed objects if you don't use extra features.  Don't feel too strongly about the naming here, just hoped there was something more specific/clearer because just googling gave me a lot of context/intro and I couldn't quickly find just a simple definition.  (Note that in testthat you can define objects in a test file outside calls to `test_that` and they can be used within the rest of the `test_that`s in the file, but won't be available in other test files.  Not saving API calls like the rds file approach, but thought I'd mention it because I didn't see this pattern yet in epidatr tests.)

dshemetov:
Sorry, the `r$results != 2` is a typo from me. Meant to be -2.

brookslogan:
todo: forward the (html-format) error message the server provides when we stop for status; there's code to borrow [here](https://github.com/cmu-delphi/delphi-epidata/blob/4d39e96b7cbf376dc9cfa273012b98db22e1225a/src/client/delphi_epidata.R#L57-L67)

This + appropriately warning on non-"success" error messages + avoiding the csv transfer format are important for the API keys rollout.  Are we in good shape to merge this soon?  I'll start checking off things from my long list I know have already been addressed; some polish stuff can also be left for later.

brookslogan:
> wrt epidata warnings: should we catch other epidata status codes? is there even documentation for epidata status codes? i have a feeling we'll have to dig through server code to find these

My guess is that, for `fetch_tbl` via `"classic"` format, we need to stop on non-1 `result`s, and warn on `result`s of 1 with non-`"success"` `message`s.  But I have not done the digging.

For `fetch_classic` if that's still in, it's less clear, if we're still returning them the result-message-epidata list.  We'd like to stop them from continuing without checking for errors/warnings, but it's also weird when they get this `response$result` thing that can only be 1 in commands that actually complete.

dshemetov:
I'm a bit overwhelmed by the volume of comments here. I have addressed many of them, don't understand some of them, need more information for others. Do you have time to sync so we don't have to go back and forth through github on this?

dshemetov:
Also, can you turn all your points into todos so I can mark them as checked off? It would be easier than writing a comment responding to all of them.

R/epidatacall.R
```diff
@@ -68,187 +58,214 @@ create_epidata_call <- function(endpoint, params, meta = NULL,
...
+#' Fetches the data
+#'
+#' `fetch` usually returns the data in tibble format, but a few the endpoints
+#' only support the JSON classic format. In that case a JSON-like nested list
+#' structure is returned instead.
+#'
+#' @rdname epidata_call
+#' @param epidata_call an instance of `epidata_call`
+#' @param fields filter fields
+#' @param disable_date_parsing disable automatic date parsing
```
brookslogan, line 110: polish: "[...] in response columns"

dshemetov:
My mental map of error handling after reading your comments looks like:

1. There's HTTP errors
2. There's HTTP messages in addition to errors? I don't even know what this looks like.
3. There's epidata messages/warnings/errors
4. There's the R messages/warnings/errors

I think I have handled 1, 3, and 4 in this PR already. I probably need to integrate [this code](https://github.com/cmu-delphi/delphi-epidata/blob/4d39e96b7cbf376dc9cfa273012b98db22e1225a/src/client/delphi_epidata.R#L57-L67) in, but I wonder what those error codes are and if they actually have helpful messages in the body.

Separately, I am not particularly concerned about having 'classic' return a status code of 1 and a 'success' message. First, it doesn't seem that confusing to have that message there. People can easily ignore it. Second, we use those fields helpfully if they DO return something other than that. And finally, I expect that 99% of users will use `fetch_tbl` anyway, so they'll never see those extra fields.

R/epidatacall.R
```diff
@@ -68,187 +58,219 @@ create_epidata_call <- function(endpoint, params, meta = NULL,
...
 
-  m <- jsonlite::fromJSON(r)
-  if ("epidata" %in% names(m) && !(epidata_call$only_supports_classic)) {
-    m$epidata <- parse_data_frame(epidata_call, m$epidata, disable_date_parsing = disable_date_parsing)
+  r <- jsonlite::fromJSON(r, simplifyDataFrame = !disable_data_frame_parsing)
+  # success is 1, no results is -2
+  if (r$result != 1 && r$result != -2) {
+    rlang::abort(paste0("epidata error: ", r$message), "epidata_error")
+  } else {
+    r
```
brookslogan, line 204: question: should we `rlang::warn` if `r$message != "success"`?
question: if yes, why not just return `r$epidata` at the end of this?  We've already propagated all errors and warnings.

question: do we even want to allow partial results?  There's no good way to get the rest of the results.  I think it'd be more helpful for that to be a hard error.

R/epidatacall.R
```diff
@@ -68,187 +58,219 @@ create_epidata_call <- function(endpoint, params, meta = NULL,
...
   if (epidata_call$only_supports_classic) {
-    rlang::abort("the endpoint only supports the classic message format, due to a non-standard behavior",
+    rlang::abort("This endpoint only supports the classic message format, due to a non-standard behavior. Use fetch_classic instead.",
       epidata_call = epidata_call,
       class = "only_supports_classic_format"
     )
   }
+
   res <- request_impl(epidata_call, "csv", fields)
   httr::stop_for_status(res)
```
brookslogan, line 230: todo: same `stop_for_status` http error messaging stuff should be here as above.  Could be refactored into its own method.

R/epidatacall.R
```diff
@@ -68,187 +58,219 @@ create_epidata_call <- function(endpoint, params, meta = NULL,
...
-    columns <- colnames(tbl)
-    for (i in seq_len(length(meta))) {
-      info <- meta[[i]]
-      if (info$name %in% columns && info$type == "epiweek") {
-        tbl[[info$name]] <- parse_api_week(tbl[[info$name]])
-      }
-    }
-  }
-  tbl
+  df <- fetch_classic(epidata_call, fields, disable_data_frame_parsing = FALSE)$epidata
```
brookslogan, line 154: todo: if we don't `rlang::warn` on non-"success" message in `fetch_classic`, it must be done here.

brookslogan:
Okay, I think I've transferred everything remaining to the relevant lines.  You might have already seen this from incorporating [this snippet](https://github.com/cmu-delphi/delphi-epidata/blob/4d39e96b7cbf376dc9cfa273012b98db22e1225a/src/client/delphi_epidata.R#L57-L67), but a server can return an http error code PLUS content.  Like on [github](https://github.com/brookslogan/repo-that-does-not-exist) in a browser, you can get a 404 plus html content.  Our API servers can do the same, and will do the same for API keys-related error codes and messages.  The default error code descriptions that `httr::stop_for_status` would show by default, e.g., something like "Too Many Requests" for http error 429 are descriptive, but don't give all the details we want, like how to sign up for a key to get a higher/no rate limit.

R/epidatacall.R
brookslogan: (and I'd also like to raise an error on truncated requests even if we decide against it in `fetch_classic`.  there's no real way to make use of the results you do get.)

R/epidatacall.R
dshemetov: See my comment below. I'd rather not have someone call fetch_classic and get the results of fetch_json. So I'd prefer to just not raise errors at all and maybe warn instead.

As for the final question, I don't think our server notifies when results are partial. But even if it does, it seems strange to let someone download 10 million rows and then to discard the download with a "partial" error.

brookslogan: I'd then want something like `fetch_json` (but without dataframeification) or the alternative name `fetch_object`, which DO stop and warn in all cases, and would advertise those over `fetch_classic`, and use them in `fetch` rather than `fetch_classic`.  Reasoning: `fetch_classic` is just asking for people to accidentally ignore errors and warnings, as we did in this very client before.  And `fetch_object` and `fetch_tbl` are more similar to each other, so they  would mix better / surprise less in `fetch()`.

Do we still have endpoints around that have a limit of 3650 rows?  Guess I have some old annoyances that might not apply now.  But why is the server doing truncation and not a hard error at 10M rows?  It's errors waiting to happen again.  I guess it's hard to know what how many rows a query will generate ahead of time, but if I were a user I'm probably not going to try to debug exactly which 10M rows I got and which I'm missing and really rely on my guesses in getting the missing rows with another (potentially very hard to write) query.  I'd dump the 10M and rewrite to fetch the entire thing in a way that wouldn't hit the limit in any individual query.

dshemetov: Re: limits. Looks like people still hitting the old `api.php` endpoint still hit the 3650 row limit, I guess for compatibility reasons 🤦. Also [the server gives a status code of 2 when it hits the limit](https://github.com/cmu-delphi/delphi-epidata/blob/52f57d6cd6d3da967c24e9d2741e727560bf3e27/src/server/_printer.py#L88-L91). People using this client, unless for some reason they choose to `with_base_url` the old php endpoint, a choice that I think we can safely ignore, will have the 10e6 limit. You're right, the partial results aren't very useable, so I am fine with throwing an error if we get a truncated message. It would suck for the user to wait a long time to only get an error back, but I don't think there's much to do about that.

Re: fetch functions. I just don't see why we need to export all these `fetch_*` functions. I thought we were aligned on this PR's goal of centralizing everything around `fetch`, but I suppose not? I would prefer to delete `fetch_json`, unexport `fetch_classic` and `fetch_tbl`, and only export `fetch`. This function will then return a tibble in 99% of use cases, will raise on HTTP errors, surface HTTP body messages as warnings, will raise on epidata errors, and in the few only_supports_classic cases, will return a raw JSON object from the API call. The three remaining only_supports_classic endpoints are either barely used or are already covered - `delphi` is [broken](https://api.covidcast.cmu.edu/epidata//delphi/?system=ec&epiweek=202006), `pvt_meta_norostat` is one of the private endpoints [untouched in years](https://delphi-org.slack.com/archives/C02U0AX8KJR/p1683837318090089), and `meta` which is already covered by `covidcast_epidata()`. The classic format is explained to the user in the docstring, so they'll know what to expect from these endpoints.

brookslogan: I thought the `fetch` idea here was originally to /default/ to tbl or classic, but also /allow/ other things.  As we discuss though, I'm okay with trying to eliminate some of the options.  But I don't think result-status-epidata without stop/warn should ever be the default output, and it stop&warn&return epidata should be an option for all endpoints.  Plus it seems a little simpler and less error-prone to have `fetch` to stop&warn in all cases, and return the epidata in all cases (turned into tbl for non-only-classic endpoints only).  The problem with not stopping is that people don't check the status&message and file bug reports, and the problem with not warning (larger) is that they may NEVER catch these and just have bad input & bad output.

R/epidatacall.R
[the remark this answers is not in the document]
dshemetov: update some of these

R/epidatacall.R
dshemetov: done

R/epidatacall.R
dshemetov: changed to raise on non-success in fetch_classic and return result$epidata

tests/testthat/test-epidatacall.R
```diff
@@ -0,0 +1,60 @@
...
+    geo_type = "state",
+    time_values = epirange("2020-06-01", "2020-08-01"),
+    geo_values = "ca,fl"
+  )
+  # Generated with
+  # epidata_call %>%
+  # fetch_debug(format_type = "classic") %>%
+  # readr::write_rds(testthat::test_path("data/test-classic.rds"))
+  mockery::stub(fetch_classic, "httr::content", readRDS(testthat::test_path("data/test-classic.rds")))
+  mockery::stub(fetch_tbl, "fetch_classic", fetch_classic)
```
brookslogan, line 15: note: just the single `stub` wasn't working, at least on my system.  See a little discussion at https://github.com/cmu-delphi/epidatr/pull/99/commits/4b6f964beb1aefb350b6bd7d540f9d33416d02b1 and https://github.com/cmu-delphi/epidatr/issues/110.

brookslogan (approved):
Looks good!  Made a few edits; please give them a sanity check / fixup.

brookslogan:
question: can we eliminate `fetch_csv` altogether?  Or is it used internally somewhere / otherwise desirable?

R/epidatacall.R
```diff
@@ -68,187 +58,213 @@ create_epidata_call <- function(endpoint, params, meta = NULL,
...
 #' @importFrom httr stop_for_status content
-#' @importFrom jsonlite fromJSON
 #' @importFrom rlang abort
-#' @return
-#' - For `fetch_json`: a `data.frame` form of the object from the parsed JSON
-#'   response
-#'
-#' @export
-fetch_json <- function(epidata_call, fields = NULL, disable_date_parsing = FALSE) {
+fetch_csv <- function(epidata_call, fields = NULL, disable_date_parsing = FALSE, disable_tibble_output = FALSE) {
```
brookslogan, line 214: question: can we eliminate `fetch_csv` altogether?  Or is it used internally somewhere / otherwise desirable?

brookslogan:
Going to go ahead and merge this, so we have warnings and the new interface for API keys rollout.

R/epidatacall.R
dshemetov: I think we can. I kept it around just so we had a test reference for the new `fetch_tbl`.

PR #103, opened 2023-05-15
update base url to reference `api.covidcast.cmu.edu/epidata`

melange396:
see https://github.com/cmu-delphi/delphi-epidata/pull/1164


dshemetov (approved):
:shipit:

PR #114, opened 2023-05-26
refactor(endpoints): remove API arg exclusivity validation

### LINKED ISSUE #92: Update issues&lags exclusivity checks to also consider `as_of`
I believe `covidcast` no longer supports `as_of` together with `issues` or `lags`.  Other endpoints might still support.  We should update the client-side checks.  I believe it's also in an Issue somewhere to detect this on the server side as well.

dshemetov:
Based on the discussion in cmu-delphi/delphi-epidata#1159, if we commit to doing the exclusivity checks in the server, then with our changes in #99, we will forward API validation errors to the user here, so no need to duplicate the work in the client.

Closes #92 (via changed design direction).

### DIFF
--- a/R/endpoints.R
+++ b/R/endpoints.R
@@ -80,16 +80,6 @@ covid_hosp_facility_lookup <- function(state = NULL, ccn = NULL, city = NULL, zi
   assert_character_param("zip", zip, len = 1, required = FALSE)
   assert_character_param("fips_code", fips_code, len = 1, required = FALSE)
 
-  if (
-    missing(state) &&
-      missing(ccn) &&
-      missing(city) &&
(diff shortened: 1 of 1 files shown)


brookslogan:
Can/should this wait until we get the server-side checks in https://github.com/cmu-delphi/delphi-epidata/issues/1159?  Not sure if there's some GitHub magic to auto-merge or auto-undraft when an issue in another repo is closed.

dshemetov:
It would be convenient to merge now for us, just to get it off our plate, but it does remove a few guard rails for the users. I would guess that these wouldn't be very common errors, so it's not a big deal, but I don't have data on that

PR #115, opened 2023-05-26
refactor(epidatacall): remove `fetch_csv` and use `testthat` mocking

### LINKED ISSUE #110: Investigate `testthat::{with,local}_mocked_bindings`
We might want to take a look at `testthat::{with,local}_mocked_bindings`, which takes lessons from `mockr` and `mockery`.  (This is not to be confused with the deprecated `testthat::with_mock`.)

Issues with status quo: while testing #99 I encountered some issues with `mockery::stub` where (a) a natural-seeming `stub` call wasn't actually correct, and (b) [adjusting to try to use `depth` didn't seem to work](https://github.com/cmu-delphi/epidatr/pull/99/commits/4b6f964beb1aefb350b6bd7d540f9d33416d02b1), though a workaround was easy, and (c) when attempting to debug some tests, I wanted to evaluate some chunks of the test in a REPL, but running `stub` has lasting effects unless packaged inside a `test_that` or `local({ <code> })`.

dshemetov:
Fixes #110 and addresses [this comment](https://github.com/cmu-delphi/epidatr/pull/99#discussion_r1201252965).


tests/testthat/test-epidatacall.R
```diff
@@ -51,10 +40,14 @@ test_that("fetch_tbl warns on non-success", {
     readRDS(testthat::test_path("data/test-classic.rds")) %>%
     jsonlite::fromJSON() %>%
     `[[<-`("message", "* This is a warning with a leading asterisk and {braces} to make sure we don't have bulleting/glue bugs.")
```
brookslogan, line 42:
```suggestion
    `[[<-`("message", artificial_warning)
```

I introduced `artificial_warning` and must have forgotten to complete the de-dupe.

tests/testthat/test-epidatacall.R
```diff
@@ -7,32 +7,21 @@ test_that("fetch_tbl", {
...
+    {
+      tbl_out <- epidata_call %>% fetch_tbl()
+      out <- epidata_call %>% fetch()
+    },
+    # RDS file generated with
+    # epidata_call %>%
+    # fetch_debug(format_type = "classic") %>%
+    # readr::write_rds(testthat::test_path("data/test-classic.rds"))
+    content = function(...) readRDS(testthat::test_path("data/test-classic.rds")),
+    .package = "httr"
```
brookslogan, line 22: note: https://testthat.r-lib.org/reference/local_mocked_bindings.html says that this approach with `.package =` is "not a great idea because it affects all code in that package, not just code in your package".  I think it's probably fine here though, since we don't expect any other calls to `httr::content` besides our own direct calls.  Just noting in case you didn't already see.  This might factor into discussion [here](https://github.com/cmu-delphi/epiprocess/issues/112).

The `jsonlite::fromJSON` mock also appears to work, although I feel less certain that we wouldn't use it indirectly via `httr` vs. in the above case.  But not sure it's worth doing anything unless we actually make a decision in `::`-related discussion.

R/epidatacall.R
```diff
@@ -334,37 +268,3 @@ request_impl <- function(epidata_call, format_type, fields = NULL) {
...
-    bool = readr::col_logical(),
-    text = readr::col_character(),
-    int = readr::col_integer(),
-    float = readr::col_double(),
-    categorical = readr::col_factor(info$categories)
-  )
-  r <- types[info$type]
-  stopifnot(!is.null(r))
-  r
-}
```
brookslogan, line 370: praise: nice catch eliminating this unused code.  Pretty exciting to see some complexity go away!

tests/testthat/test-epidatacall.R
brookslogan: question: do we even need this test anymore?  IIRC the old version was testing that we obtained the same content from parsing classic as from parsing csv.  But now both of these are based on the classic transfer format, and `fetch` even directly calls `fetch_tbl`.

brookslogan (approved):
Looks good.  Have some non-blocking discussion and suggested related/nearby cleanup.

R/epidatacall.R
dshemetov: it's so cathartic deleting code :D

tests/testthat/test-epidatacall.R
dshemetov: hm, yea this test isn't doing much. we should probably replace it with a parse function.

good point about mocking, definitely worth keeping in mind in the future.

tests/testthat/test-epidatacall.R
```diff
@@ -7,32 +7,21 @@ test_that("fetch_tbl", {
...
-  # format) from the API, uses jsonlite::fromJSON to convert the underlying data
-  # to a data.frame, and finally applies parse_data_frame is used to do the data
-  # type coersion specified by the epidata_call metadata.
-  # 2) fetch_csv requests the csv format from the API, then uses readr::read_csv
-  # to get a data.frame, and has its own methods to enforce data types.
-  tbl_out <- epidata_call %>% fetch_tbl()
-  csv_out <- epidata_call %>% fetch_csv()
-  expect_identical(tbl_out, csv_out)
+  # This test compares the output of a tibble using fetch and fetch_tbl.
+  with_mocked_bindings(
```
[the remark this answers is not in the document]
dshemetov: thanks for pointing this out. the problem was that I was mocking httr::content, which parse the response, but i didnt mock epidatr::request_impl, which was still making a request (even though the request wasn't being used).

PR #125, opened 2023-06-16
Internals docs

### LINKED ISSUE #108: Consider `@noRd` for `assert_*` and any other non-exported functions
Even though these functions are not exported, they still will show up in the "Reference" with the exported functions.  Assuming this is unintentional, we should add `@noRd` (no .Rd / compiled doc at all) or `@keywords internal` (documentation is compiled, but not listed in the "Reference" index); people seem to favor the first approach.

dsweber2:
closes #108 
In the process of doing this, I noticed that the [epidata_call](https://cmu-delphi.github.io/epidatr/reference/epidata_call.html) docs are in a weird state. On the [reference page](https://cmu-delphi.github.io/epidatr/reference/index.html), there's a mass of functions all together `create_epidata_call() fetch() fetch_tbl() fetch_classic() request_url() with_base_url() ` which point at that page. Several of these are internal only, and some were actually moved from external to strictly internal. The second commit in the pull request may deal with that

### DIFF
--- a/R/check.R
+++ b/R/check.R
@@ -1,5 +1,6 @@
 #' Allows character vectors
 #' @importFrom checkmate assert_character assert_integerish
+#' @keywords internal
 assert_character_param <- function(name, value, len = NULL, required = TRUE) {
   null.ok <- !required
   assert_integerish(len, null.ok = TRUE, .var.name = "len")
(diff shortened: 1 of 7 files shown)


dsweber2:
actually confirmed that the last one displays `create_epidata_call() fetch() with_base_url()` in the reference section

dshemetov (approved):
I wish we had a way to preview these in PRs...

Otherwise, these specific changes look good. Since this isn't a final doc draft, I'm just going to let this go through and we can see what the docs look like as the CI builds them

PR #126, opened 2023-06-26
docs: actual example for `pvt_norostat`

dsweber2:
`pvt_norostat`'s locations are a bit unusual. This gives at least one functional example. Ideally we'd also point at a full list of the locations, but as per https://github.com/cmu-delphi/delphi-epidata/issues/1209 there isn't such a public list atm


dshemetov (approved):

brookslogan:
Possible locations are dynamic as the set of FluSurv-NET participating states changes.  Location possibilities can be fetched via `pvt_meta_norostat`.  Filling out a bit more info in the linked issue.

Issue #128, opened 2023-06-30
Dealing with large requests


dsweber2:
Doing 
```
chng_outpatient <- epidatr::covidcast(
  data_source = "chng", signals = "smoothed_adj_outpatient_covid",
  geo_type = "state",
  time_type = "day",
  time_values = "*",
  geo_values = "*",
  issues = "*"
)
```
resulted in a `504`, probably because it took too long to get back; it worked fine for restricting `issues` to a month interval. Ways to address this:
1. Expose whatever [means](https://httr.r-lib.org/reference/timeout.html) `httr` has of increasing the wait time as a user input
2. document any already existing way of doing this
3. Split `"*"` requests into reasonable sized subsets and then do joins after to return what they asked for.

PR #132, opened 2023-07-05
Extreme requests

dsweber2:
Adds two parameters to several functions:
- `timeout_seconds` gives the maximum time before a connection will time out. Default is 30 seconds, which matches the value returned in the call below (can't find documentation for httr's default, since it clearly isn't using [libcurl's](https://curl.se/libcurl/c/CURLOPT_TIMEOUT.html)). Originally motivated by
  ```
  chng_outpatient <- epidatr::covidcast(
    data_source = "chng", signals = "smoothed_adj_outpatient_covid",
    geo_type = "state",
    time_type = "day",
    time_values = "*",
    geo_values = "*",
    issues = epirange("2020-11-16","2021-03-16")
  ) %>% fetch()
  ```
  which returns a timeout message, trying 3 times. A working version in this branch is
  ```
  chng_outpatient <- epidatr::covidcast(
    data_source = "chng", signals = "smoothed_adj_outpatient_covid",
    geo_type = "state",
    time_type = "day",
    time_values = "*",
    geo_values = "*",
    issues = epirange("2020-11-16","2021-03-16")
  ) %>% fetch(timeout_seconds = 2.5 * 60)
  ```
  I'm not sure there's a nice way to add a unit test for something like this however. Perhaps `httptest` has something.
- `return_empty`, which handles the case of no returned results whatsoever. Defaults to `FALSE`, which is the current value. Even if its true it throws a warning. We may want to consider making `TRUE` the default. I debated adding headers to empty columns, but it ended up being more awkward to use on the other end.
  Should be relatively easy to add tests, but I haven't yet. It does work in the pipeline that motivated it (breaking the above into months, some of which `chng` has no data).

### DIFF
--- a/R/covidcast.R
+++ b/R/covidcast.R
@@ -100,15 +100,16 @@ print.covidcast_data_source <- function(source, ...) {
 #' print(smoothed_cli)
 #' df <- smoothed_cli$call("nation", "us", epirange(20210405, 20210410))
 #' @param base_url optional alternative API base url
+#' @param timeout_seconds the maximum amount of time to wait for a response
 #' @importFrom httr stop_for_status content http_type
 #' @importFrom jsonlite fromJSON
 #' @importFrom xml2 read_html xml_find_all xml_text
 #' @return an instance of covidcast_epidata
 #'
 #' @export
-covidcast_epidata <- function(base_url = global_base_url) {
+covidcast_epidata <- function(base_url = global_base_url, timeout_seconds = 30) {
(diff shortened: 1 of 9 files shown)


dshemetov (approved):
looks good!

- tests for timeout_seconds not necessary, since the codepath is still the same
- a single test for return_empty might be good, but up to you, purely a code cov thing

dshemetov:
#128

PR #149, opened 2023-08-16
Docs review

capnrefsmmat:
Here's an initial batch of changes from reading through the package documentation this morning, and running `devtools::check(remote=TRUE)`.  Each commit is small and the message should explain the purpose, so it may be easier to review each commit separately.


R/covidcast.R
```diff
@@ -32,10 +32,10 @@ parse_signal <- function(signal, base_url) {
 }
 
 #' @export
-print.covidcast_data_signal <- function(signal, ...) {
```
dsweber2, line 34: Not sure why you're preferring a more generic variable name?

R/epidatr-package.R
dsweber2, line 1: What a peculiar bit of boilerplate. I'll definitely be taking your word for it

R/endpoints.R
dsweber2, line 1: oh, there's apparently another `data_source`->`source` we missed, the stop is printing the wrong error message (line 688)https://github.com/cmu-delphi/epidatr/blob/300cfcbe59f5ccce41f1f0366cb3fde91759f9a4/R/endpoints.R#L688

vignettes/epidatr.Rmd
```diff
@@ -14,18 +14,49 @@ library(epidatr)
 library(dplyr)
 ```
 
+The epidatr package provides access to all the endpoints of the [Delphi Epidata
```
dsweber2, line 17: definitely needed to be less sparse

dsweber2 (commented):
Glad to see this! Only one code style thing I might disagree with, otherwise lots of necessary fixes.

R/covidcast.R
capnrefsmmat: We're required to use `x`, because that's how `print()` is defined in base R:

```
> print
function (x, ...) 
UseMethod("print")
```

See https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Generic-functions-and-methods

R/epidatr-package.R
capnrefsmmat: See https://roxygen2.r-lib.org/articles/rd-other.html#packages

dshemetov: Was also curious, thanks!

capnrefsmmat:
Pushed a fix for the error message saying `data_source` when it should say `source`, plus a couple other fixes. If you can review and merge this before the meeting this week, that'd be great, since it'll let the review start from here.

vignettes/epidatr.Rmd
```diff
@@ -14,32 +14,81 @@ library(epidatr)
...
+list for you. Note that some signal names have dashes in them, so to access them
+we rely on the backtick operator:
+
+```{r}
 epidata$signals$`fb-survey:smoothed_cli`
 ```
 
-From there, you can construct a call analogous to the one above, but using the API object instead of the functions directly.
+These objects can be used directly to fetch data, without requiring us to use
+the `covidcast()` function. Simple use the `$call` attribute of the object:
```
dshemetov, line 91:
```suggestion
the `covidcast()` function. Simply use the `$call` attribute of the object:
```

PR #150, opened 2023-08-22
Fix minor typo in documentation

willtownes:
"can we" changed to "can be"


dshemetov:
Thanks Will! What is this .rproj file about? Is it an RStudio thing? Is it common to include such settings in a package?

willtownes:
yes, the rproj file is just a text file with some settings used by Rstudio. Feel free to omit it from the merge, I was in a hurry yesterday and didn't think to do the pull request based off of the earlier commit instead of the latest. If you do include it I don't think it will hurt anything when submitting to CRAN.

dshemetov (approved):

PR #155, opened 2023-08-25
Djm/endpointfun

### LINKED ISSUE #147: `as.data.frame()` methods don't match the generics
`R CMD check` reports:

```
W  checking S3 generic/method consistency (660ms)
   as.data.frame:
     function(x, row.names, optional, ...)
   as.data.frame.covidcast_data_signal_list:
     function(signals, ...)
   
   as.data.frame:
     function(x, row.names, optional, ...)
   as.data.frame.covidcast_data_source_list:
     function(sources, ...)
   See section ‘Generic functions and methods’ in the ‘Writing R
   Extensions’ manual.
```

These methods need to accept exactly the same argument names as the base `as.data.frame()`, plus optionally any additional ones. I've fixed a couple other generics in my branch, but these are trickier because we need to decide what to do with `row.names` and `optional`.

### LINKED ISSUE #148: Make CI fail on check warnings, not just errors
Currently the CI is set to allow check warnings: https://github.com/cmu-delphi/epidatr/blob/f9166b8ea4243d5a0ef469a13af80dca83df9b02/.github/workflows/ci.yml#L34-L39

However, CRAN generally does not accept packages with check warnings unless you have a very good reason, so we should change this to reject warnings as well -- as soon as we fix our existing warnings, such as in #147.

### LINKED ISSUE #154: Add @keywords
per suggestion by @dajmcdon:

> To view “all possible endpoints”, we could add @keyword tags to each of their documentation files. This has 2 benefits. One is that this can be easily (and quickly) searched. So we could write a small helper, say, available_endpoints(), that would just list them all. If we add more later, we just remember to add the same keyword to the epicall function. The second benefit is that the _pkgdown.yml reference index can contain has_keyword(…), so we don’t need to list all the endpoints individually. Since this (and likely lots of other parts of the documentation) are shared across all signals, we could even extract anything shared to a single doc. I can probably do this as a PR if others think it would be worth having.

dajmcdon:
1. adds a `avail_endpoints()` to list all available endpoints
2. adjusts `_pkgdown.yml` to simplify the reference build
3. removes "Fetch" as the first word in the title of the endpoint man files (it doesn't fetch, and is a bit redundant even if it does)
4. The `Meta/` directory should not be under version control.
5. Moved the `development.rmd` vignette to `DEVELOPMENT.md`, don't build it.
6. Use `&&` instead of `&` in `covidcast.R` (lintr wants this, and it's better for if/else statements)
7. Wrap `epirange()` in the FluView documentation.
8. Fix S3 method/generic consistency in `as.data.frame.xxx()` funs (`covidcast.R`)

closes #154
closes #147 
closes #148


R/avail_endpoints.R
```diff
@@ -0,0 +1,20 @@
+#' List all available endpoints
+#'
+#' @return A [`tibble::tibble`] with 2 columns. `Endpoint` contains the function
+#'   for accessing the Delphi Epidata API endpoint along with a `Description`.
+#' @export
+#'
+#' @examples
+#' endps <- avail_endpoints()
+#' print(eps, n = nrow(endps))
```
capnrefsmmat, line 9:
```suggestion
#' print(endps, n = nrow(endps))
```

PR #156, opened 2023-08-25
refactor: `fetch` by default

### LINKED ISSUE #82: `with_base_url` sounds like it might follow the withr/httr/httptest/testthat/etc. pattern but doesn't
Some R packages have a pattern for `with_*` functions where they will try to run some code with some setup and mandatory cleanup, similar to the Python `with` statement.  `with_base_url` sounds very much like `withr::with_options`, etc.  It'd be nice to either follow this with-statement-like pattern or find another name for this function.

### LINKED ISSUE #119: Line-wrap warning and error messages from the API
Warnings and error messages forwarded from the API aren't line-wrapped and could look better.

``` r
library(epidatr)
issue_data =
  covidcast(
    data_source = "hhs", signal = "confirmed_admissions_influenza_1d",
    geo_type = "state", time_type = "day",
    geo_values = c("az","ga"),
    # all `time_value`s:
    time_values = epirange(20221201, 20221231),
    # all issues through the analysis issue
    issues = epirange(20230101, 20230102)
  ) %>%
  fetch()
#> Warning: No API key found. You will be limited to non-complex queries and encounter rate
#> limits if you proceed.
#> To avoid this, you can get your key [here] and then:
#> ℹ set the environment variable DELPHI_EPIDATA_KEY
#> ℹ set the option 'delphi.epidata.key'
#> 
#> To save your key for later sessions (and hide it from your code), you can edit
#> your .Renviron file with:
#> ℹ usethis::edit_r_environ()
#> This warning is displayed once every 8 hours.
#> Warning: epidata warning: This request exceeded the anonymous limit on selected multiples, which will be enforced starting 2023-06-21. To be exempt from this limit, authenticate your requests with a free API key, now available at https://api.delphi.cmu.edu/epidata/admin/registration_form. A temporary public key `TEMP-API-KEY-EXPIRES-2023-06-28` is available for use between now and 2023-06-28 to give you time to register or adapt your requests without this message continuing to break your systems.
```

<sup>Created on 2023-06-05 with [reprex v2.0.2](https://reprex.tidyverse.org) but altered to show that we're not actually wrapping the message.</sup>

We might be able to use `cli` for this, but [need to be careful](https://rlang.r-lib.org/reference/topic-condition-formatting.html#transitioning-from-abort-to-cli-abort-) not to use the message directly as a cli format string.  (I vaguely recall fiddling with this and using the brace syntax for gluing in a string and it not doing the line wrapping though, but my memory could be bad.  There might be some special cli/glue thing to enable the line wrap on interpolated strings.  Or we have some internal helpers in epiprocess for line wrapping, but that's sort of backwards progress.)

### LINKED ISSUE #144: Consider `fetch`ing by default?

dshemetov:
closes #144
closes #119 

- `fetch_args_list()` has these arguments, open to better/clearer naming suggests

```r
#' @param fields a list of epidata fields to return, or NULL to return all
#'  fields (default) e.g. c("time_value", "value") to return only the
#' time_value and value fields or c("-direction") to return everything except
#' the direction field
#' @param disable_date_parsing disable automatic date parsing
#' @param disable_data_frame_parsing disable automatic conversion to data frame; this
#'   is only supported by endpoints that only support the 'classic' format (non-tabular)
#' @param return_empty boolean that allows returning an empty tibble if there is no data
#' @param timeout_seconds the maximum amount of time to wait for a response
#' @param base_url base URL to use
#' @param make_call boolean that allows skipping the call to the API and instead
#'  returns the `epidata_call` object (useful for debugging)
#' @param debug boolean that allows returning the raw response from the API
#' @param format_type the format to request from the API, one of classic, json, csv; this
#'  is only used by `fetch_debug`
```

- all endpoints now fetch by default
- with_base_url has been absorbed into fetch args and made internal which should close #82
- added automated ci actions (automatically documents, styles, and lint should now actually complain properly)

TODO:
- [x] look over the vignette text to make sure it still reads right


dsweber2:
Generally pretty happy with this, made a number of small docs changes. To keep the format of the flags the same, maybe `make_call` should be `dry_run` or something like that?

I separated `fetch_args_list` into a separate reference page, because all the parameters of `create_epidata_call`, `fetch` and `fetch_args_list` were mixed together.

PR #178, opened 2023-09-15
ci: remove release helper

dshemetov:
Since we have switched to a 1 trunk repo with `dev`, the release process is a lot less involved - just hit the new release button in GH. Remove the automation set to trigger based on `main`.


dsweber2 (approved):
:shipit:

PR #189, opened 2023-09-29
fix: cache size integer overflow

### LINKED ISSUE #188: Can't use a large cache size limit
``` r
epidatr::set_cache(max_size = 10000L)
#> Warning in cachem::cache_disk(dir = cache_dir, max_size = as.integer(max_size *
#> : NAs introduced by coercion to integer range
```

<sup>Created on 2023-09-28 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

I believe R always uses 32-bit integers, so `10000L * 1024^2` is too large.

dshemetov:
fix #188 

since the default for [max_size in cache_disk](https://cachem.r-lib.org/reference/cache_disk.html) is `1024 * 1024^2`, a double, that argument can eat doubles, so remove the `as.integer` cast that's overflowing

### DIFF
--- a/R/cache.R
+++ b/R/cache.R
@@ -163,7 +163,7 @@ set_cache <- function(cache_dir = NULL,
   } else if (cache_exists) {
     cache_environ$epidatr_cache <- cachem::cache_disk(
       dir = cache_dir,
-      max_size = as.integer(max_size * 1024^2),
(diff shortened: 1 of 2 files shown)


dsweber2 (approved):
who doesn't use 64bit integers...

PR #190, opened 2023-09-30
lint: update lintr

dshemetov:


dsweber2 (approved):
oh, right, I suppose that linting the `renv` could cause problems. Surprised we have a `venv`

PR #191, opened 2023-09-30
style: use glue for line-wrapping in cache warnings

dshemetov:


dsweber2 (approved):

R/cache.R
```diff
@@ -252,12 +252,12 @@ cache_epidata_call <- function(epidata_call, fetch_args = fetch_args_list()) {
...
-          "1. disabling the cache for this session with `disable_cache` or permanently with environmental ",
-          "variable `EPIDATR_USE_CACHE=FALSE`\n",
-          "2. setting `EPIDATR_CACHE_MAX_AGE_DAYS={Sys.getenv('EPIDATR_CACHE_MAX_AGE_DAYS', unset = 1)}`",
-          " to e.g. `3/24` (3 hours)."
+          "Using cached results with `as_of` within the past week (or the future!).
+          This will likely result in an invalid cache. Consider",
+          "i" = "disabling the cache for this session with `disable_cache` or
+          permanently with environmental variable `EPIDATR_USE_CACHE=FALSE`",
+          "i" = "setting `EPIDATR_CACHE_MAX_AGE_DAYS={Sys.getenv('EPIDATR_CACHE_MAX_AGE_DAYS
+          ', unset = 1)}` to e.g. `3/24` (3 hours)."
```
brookslogan, line 260: todo: This rendered as `ℹ setting `EPIDATR_CACHE_MAX_AGE_DAYS=1` to e.g. `3/24` (3 hours).` for me.  Tweak to remove or reword the `=1`.

R/cache.R
```diff
@@ -267,8 +267,8 @@ cache_epidata_call <- function(epidata_call, fetch_args = fetch_args_list()) {
     if (!is.key_missing(cached)) {
       cli::cli_warn(
```
brookslogan, line 268: Aside: could we make this a `cli_inform` or maybe just silent (or silent + opt-in verbose cache setting?).

R/cache.R
```diff
@@ -252,12 +252,12 @@ cache_epidata_call <- function(epidata_call, fetch_args = fetch_args_list()) {
...
         c(
-          "using cached results with `as_of` within the past week (or the future!).",
-          "This will likely result in an invalid cache. Consider\n",
-          "1. disabling the cache for this session with `disable_cache` or permanently with environmental ",
-          "variable `EPIDATR_USE_CACHE=FALSE`\n",
-          "2. setting `EPIDATR_CACHE_MAX_AGE_DAYS={Sys.getenv('EPIDATR_CACHE_MAX_AGE_DAYS', unset = 1)}`",
-          " to e.g. `3/24` (3 hours)."
+          "Using cached results with `as_of` within the past week (or the future!).
+          This will likely result in an invalid cache. Consider",
+          "i" = "disabling the cache for this session with `disable_cache` or
```
brookslogan, line 257: comment: "i" = is "info item", not numbered list, if that was what you were expecting.  Here it might look prettier to use "*" for a bulleted list.  (Not sure how to do an actual enumeration, but here just an unordered list seems more suitable.)

brookslogan (changes requested):
Very minor tweak + requests.

R/cache.R
dsweber2: I made a standalone issue for this, probably a good idea

R/cache.R
dsweber2: I think that was intended, to show the current value? though it does look like github is treating the ` strangely.

dshemetov: maybe we reword it for clarity?

```
setting `EPIDATR_CACHE_MAX_AGE_DAYS` (currently {Sys.getenv('EPIDATR_CACHE_MAX_AGE_DAYS'
, unset = 1)}) to e.g. `3/24` (3 hours)
```

brookslogan: it might also benefit from some extra explanation what it's doing, making sure that they receive any new versions of versions quickly (but with better wording)

PR #195, opened 2023-10-06
fix: actual license details

dsweber2:
Following R conventions on the owners, as its ambiguous which orgs exactly would hold. Also adding the particular rights claimed...

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -22,3 +22,4 @@
 ^.env$
 ^README\.Rmd$
 ^CRAN-SUBMISSION$
+^LICENSE\.md$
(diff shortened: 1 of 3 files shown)


brookslogan (approved):
Diffs look right.  Any idea why macos check is failing?  Is that a preexisting or intermittent thing?

dsweber2:
```
   Error: 
  ! error in pak subprocess
  Caused by error in `file(con, "rb")`:
  ! cannot open the connection
```
:thinking: not sure this is an us problem... I don't remember it happening before

dshemetov (commented):
Why are there two license files (probably R pkgdown needs the Markdown?). Why are they different?

dsweber2:
It's what `usethis::use_mit_license()` generates, for whatever reason, so I just went with that

PR #207, opened 2023-11-04
Miscellaneous documentation improvements

capnrefsmmat:
See commit messages for descriptions of the changes. Basically I just randomly browsed the docs, saw a few things to improve, and improved them.

### DIFF
--- a/R/avail_endpoints.R
+++ b/R/avail_endpoints.R
@@ -1,11 +1,12 @@
-#' List all available endpoints.
+#' List all available Epidata API endpoints
 #'
-#' @description
-#' A function that prints a tibble with two columns: `Endpoint` contains the
-#' function for accessing the Delphi Epidata API endpoint along with a
-#' `Description`.
+#' Fetches a data frame of all Epidata API endpoints that can be accessed using
+#' this package, with a brief description.
 #'
-#' @return A [`tibble::tibble`].
+#' @return A [`tibble::tibble`] of endpoints, with two columns:
+#'   \item{Endpoint}{Name of the function for accessing this API endpoint.}
+#'   \item{Description}{One-sentence description of the data available at the
+#'   endpoint.}
 #' @export
 #' @importFrom utils help.search
 #'
--- a/R/avail_endpoints.R
+++ b/R/avail_endpoints.R
@@ -1,5 +1,6 @@
 #' List all available Epidata API endpoints
 #'
+#' @description
 #' Fetches a data frame of all Epidata API endpoints that can be accessed using
 #' this package, with a brief description.
 #'
--- a/R/epidatacall.R
+++ b/R/epidatacall.R
@@ -103,42 +103,38 @@ print.epidata_call <- function(x, ...) {
   ))
 }
 
-#' Customize fetch settings
+#' Set custom API request parameters
 #'
-#' @rdname fetch_args_list
-#' @aliases fetch_args
-#'
-#' @description
-#' A constructor for `fetch_args` objects, which are used to pass arguments to
-#' the `fetch` function.
+#' Used to specify custom options when making API requests, such as to set
+#' timeouts or change data formats. These options are used by `fetch()` when it
+#' makes calls to the Epidata API.
 #'
 #' @param ... not used for values, forces later arguments to bind by name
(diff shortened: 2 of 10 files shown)


R/avail_endpoints.R
```diff
@@ -1,11 +1,12 @@
-#' List all available endpoints.
+#' List all available Epidata API endpoints
 #'
-#' @description
-#' A function that prints a tibble with two columns: `Endpoint` contains the
-#' function for accessing the Delphi Epidata API endpoint along with a
-#' `Description`.
+#' Fetches a data frame of all Epidata API endpoints that can be accessed using
```
dsweber2, line 3: Isn't this missing a `@description`? Otherwise seems good

R/epidatacall.R
```diff
@@ -103,42 +103,38 @@ print.epidata_call <- function(x, ...) {
...
-#'  is only used by `fetch_debug`, and by default is `"json"`
-#' @return
-#' - For `fetch_args_list`: a `fetch_args` object
+#' @param fields a list of epidata fields to return, or `NULL` to return all
+#'   fields (default). e.g. `c("time_value", "value")` to return only the
+#'   `time_value` and `value` fields or `c("-direction")` to return everything
+#'   except the direction field
+#' @param disable_date_parsing disable automatic date parsing; by default
+#'   `FALSE`
+#' @param disable_data_frame_parsing disable automatic conversion to data frame;
```
dsweber2, line 119: Just to confirm, linebreaks in roxygen only show up in the code itself, and not when the documentation is displayed, right? either way, not sure how this was at 87; maybe we were wrapping at 100?

R/epidatacall.R
```diff
@@ -103,42 +103,38 @@ print.epidata_call <- function(x, ...) {
...
+#'   this is only supported by endpoints that only support the 'classic' format
+#'   (non-tabular). by default `FALSE`
```
dshemetov, line 121:
```suggestion
#'   (non-tabular); by default `FALSE`
```

dshemetov (approved):
Minor punctuation ocd, but otherwise LGTM, reads much better with your changes Alex!

dsweber2 (approved):
I'm a fan of being consistent about `@description`, but otherwise seems good; thanks for these!

R/epidatacall.R
dsweber2: is a multi-semicolon a writing smell? :thinking:

R/avail_endpoints.R
capnrefsmmat: Everything after the first line/paragraph is automatically in the Description section

R/epidatacall.R
capnrefsmmat: Actually, this raises a question: do we need `@param` to list the defaults when they're already shown in the argument list? roxygen and R show the argument list, with defaults, at the top of the help. It's telling that a few of these were wrong, showing the flaw in repeating information in different places

R/epidatacall.R
capnrefsmmat: Right, this all gets automatically reformatted when displayed. roxygen makes this a nice HTML list, while R I think automatically re-wraps lines based on your terminal width

R/epidatacall.R
dshemetov: That's a fair point, looking around other packages, like [dplyr](https://dplyr.tidyverse.org/reference/filter.html), they tend to not mention defaults in the @param field.

R/avail_endpoints.R
dsweber2: For my sanity, I'd rather include a `@description` if you don't mind. I've run into gotchas where the second paragraph automatically went into the details section, and so would rather just be explicit about sections.

capnrefsmmat:
I've pushed some fixes based on the reviews. Also I switched to using `match.arg()` for an argument that can take on one of three options, because that's how all of base R does it, and it makes the available options visible in the argument list. (See, for example, the `alternative` argument to `t.test()`.)

dshemetov:
Thanks again Alex. Also happy to learn about `math.arg()`, was wondering why people referred to those as choice arguments and not vector arguments 🙃

PR #213, opened 2023-11-10
Deploy docs website on push/pull to `dev`, since we don't have a `main` branch

nmdefries:
Website wasn't updating because workflow never triggered.


dshemetov (approved):
thank you!

PR #215, opened 2023-11-13
Support public and dev docs site

nmdefries:
https://github.com/cmu-delphi/epidatr/pull/213 keeps the doc website updated, but it matches `dev` rather than any CRAN release. If people go from CRAN to our documentation, they're going to see docs for a version they don't have. However, we'd also like to be able to see a preview of the `dev` version of the docs.

`pkgdown` lets you build doc websites from [both releases and development branches](https://pkgdown.r-lib.org/reference/build_site.html#development-mode). This PR turns on that feature, and publishes the doc site on push/pull to both `main` (reinstated to commit https://github.com/cmu-delphi/epidatr/commit/0adb796a54aee22cc54f9e1d0a724492892267b9, when [`CRAN-SUBMISSION`](https://github.com/cmu-delphi/epidatr/blob/dev/CRAN-SUBMISSION) was changed last) and `dev`. The `dev` site will be available at https://cmu-delphi.github.io/epidatr/dev; main/public at https://cmu-delphi.github.io/epidatr.

The dev website must be build from a version with 4 components (e.g. `1.1.1.1`). I've added `.9000` to the version number on `dev` to meet that requirement, but we'll need to remove it as part of the release process, since `main` needs to have a 3-part version number.

### DIFF
--- a/.github/workflows/pkgdown.yaml
+++ b/.github/workflows/pkgdown.yaml
@@ -5,9 +5,9 @@
 # update the documentation web site on pushes to `dev` branch.
 on:
   push:
-    branches: [dev]
+    branches: [main, dev]
   pull_request:
-    branches: [dev]
+    branches: [main, dev]
   release:
     types: [published]
   workflow_dispatch:
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: epidatr
 Type: Package
 Title: Client for Delphi's 'Epidata' API
-Version: 1.1.0
+Version: 1.1.0.9000
 Date: 2023-09-11
 Authors@R: 
   c(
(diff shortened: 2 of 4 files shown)


dsweber2 (approved):
oh, good to know! Needing a 4 point versioning system seems like wild overkill; maybe we could use [this pkgdown](https://github.com/insightsengineering/r-pkgdown-multiversion) instead? Up to you as the one actually implementing this.

nmdefries:
Given that the pkgdown action seems pretty unpopular/unknown, let's go with the existing implementation for now. We'll see if it makes the release process too burdensome when that happens next. (That is also manual at the moment; an automated or semi-automated release process would be able to handle the 3<->4 component version switching by itself.)

PR #217, opened 2023-11-16
refactor+docs: set_api_key

### LINKED ISSUE #181: Figure out better API key management
See the [tooling notes for Sept 19th](https://docs.google.com/document/d/15e5Fw4z3qFTntZKg3ModLCHV-z2-bWZec1zIr8rWjZ0/edit#heading=h.xwx1v0352dpw)

dshemetov:
closes #181 

Changes:
- `get_api_key()` only reads from environment variable now, no R options
- `save_api_key()` now instructs the user to write their key and dynamically opens `~/.Renviron` or `proj_root/.Renviron`
- update documentation to reflect this


R/auth.R
```diff
@@ -10,43 +10,27 @@
...
-#' are not preserved between sessions, so `set_api_key()` must be run every time
-#' you open R. Alternately, you can have R set the option at startup by adding
-#' it to your `.Rprofile`; see `?Startup` for a description of `Rprofile` files
-#' and where they can be placed.
+#' API keys are strings read from the environment variable `DELPHI_EPIDATA_KEY`.
+#' We recommend setting your key with `set_api_key()`, which will modify your
+#' `.Renviron` file and will persist your key across R sessions (see `?Startup`
+#' for a description of `.Renviron` files). Alternatively, you can modify the
+#' environment variable in the shell or with `Sys.setenv()`, but these will not
+#' persist across sessions.
```
brookslogan, line 18: From a skim, it looks like after `set_api_key()` will not set the API key for the current session, only save to the `.Renviron`.  Can this be named `save_api_key()` or maybe also set the key for the current session?

brookslogan: (Or clarify this in the docs)

dshemetov: Good point, I can rename

dshemetov: Some points about setting key automatically:

- don't know what the person wrote in .Renviron
- can't trigger an .Renviron reload with restarting R
- if the key is passed as an argument, is it safe to manually edit `.Renviron` for the user with `write(paste("DELPHI_EPIDATA_KEY=", key), file=environ_path, append=TRUE)` ?
- if the key is passed as an argument, the key will now be in the user's Rhistory

R/auth.R
```diff
@@ -64,8 +48,21 @@ get_api_key <- function() {
...
 #' @export
-set_api_key <- function(key) {
-  options(delphi.epidata.key = key)
+save_api_key <- function() {
+  cli::cli_inform(
+    c(
+      "i" = "This function will open your .Renviron file in a text editor. You will need to
+      write DELPHI_EPIDATA_KEY=yourkeyhere in the file and save it. If the editor
+      does not open, you will need to edit the file manually.",
+      "i" = "Press any key to continue."
```
capnrefsmmat, line 58:
```suggestion
      "i" = "This function will open your {.code .Renviron} file in a text editor. You will need to
      write {.code DELPHI_EPIDATA_KEY=yourkeyhere} in the file and save it. If the editor
      does not open, you will need to edit the file manually.",
      "i" = "Press any key to continue."
```

for formatting in rich-text terminals, like RStudio

R/auth.R
```diff
@@ -64,8 +48,21 @@ get_api_key <- function() {
...
+    ),
+  )
+  readline()
+
+  if (file.exists(usethis::proj_path(".Renviron"))) {
+    usethis::edit_r_environ(scope = "project")
```
capnrefsmmat, line 64: Should we print a message after each of these, saying which .Renviron was successfully edited? E.g. after project level, say "Editing project-level Renviron" and show the path to the project, so the user knows they only edited for that project

capnrefsmmat (commented):
Just a couple ideas

R/auth.R
dshemetov: `usethis::edit_r_environ` already does this when it returns!

dshemetov:
```
r$> usethis::edit_r_environ(scope="project")
✔ Setting active project to '/home/dshemeto/repos/covid-hosp-forecast'
• Modify '.Renviron'
• Restart R for changes to take effect

r$> usethis::edit_r_environ(scope="user")
• Modify '/home/dshemeto/.Renviron'
• Restart R for changes to take effect
```

dshemetov:
Might be good to go.

R/auth.R
capnrefsmmat: Very nice, great

R/auth.R
```diff
@@ -10,43 +10,27 @@
...
-#' are not preserved between sessions, so `set_api_key()` must be run every time
-#' you open R. Alternately, you can have R set the option at startup by adding
-#' it to your `.Rprofile`; see `?Startup` for a description of `Rprofile` files
-#' and where they can be placed.
+#' API keys are strings read from the environment variable `DELPHI_EPIDATA_KEY`.
+#' We recommend setting your key with `save_api_key()`, which will modify your
+#' `.Renviron` file and will persist your key across R sessions (see `?Startup`
+#' for a description of `.Renviron` files). Alternatively, you can modify the
+#' environment variable in the shell or with `Sys.setenv()`, but these will not
+#' persist across sessions.
```
brookslogan, line 18:
```suggestion
#' We recommend setting your key with `save_api_key()`, which will modify an
#' applicable `.Renviron` file, which will be read in automatically when you
#' start future R sessions (see [`?Startup`][base::Startup] for details on
#' `.Renviron` files). Alternatively, you can modify the environment variable at
#' the command line before/while launching R, or inside an R session with
#' [`Sys.setenv()`], but these will not persist across sessions.
```

R/auth.R
```diff
@@ -64,8 +48,29 @@ get_api_key <- function() {
...
 #' @rdname get_api_key
-#' @param key API key to use for future requests
 #' @export
-set_api_key <- function(key) {
-  options(delphi.epidata.key = key)
+save_api_key <- function() {
+  cli::cli_inform(
+    c(
+      "i" = "This function will open your {.code .Renviron} file in a text editor. You will need to
+      write {.code DELPHI_EPIDATA_KEY=yourkeyhere} in the file and save it. If the editor
```
brookslogan, line 56:
```suggestion
      write {.code DELPHI_EPIDATA_KEY=yourkeyhere} (without quotes) in the file and save it. If the editor
```

R/auth.R
```diff
@@ -64,8 +48,29 @@ get_api_key <- function() {
...
+      does not open, you will need to edit the file manually.",
+      "i" = "Press any key to continue."
```
brookslogan, line 58:
```suggestion
      "i" = "Press enter to continue."
```

brookslogan (approved):
Looks good.  Made a few suggestions.  There are still some references to "set" in the text; some definitely should still be "set", but not sure if others should be "save" or something else; I didn't see any place where I was convinced just "save" would be better though.

brookslogan:
One thing that would be nice, but I imagine might actually be stupidly tricky, is to first tell the user what file is being modified, before they press enter to confirm.

R/auth.R
brookslogan: GitHub's just randomly not highlighting changes for me in some places.  This reworded "your", "persist your key across", "for a description of", and "in the shell", plus added links to `?Startup` and `Sys.setenv()`.

dshemetov: Btw, `?Startup` and `Sys.setenv()` is auto-linked by roxygen (see [here](https://cmu-delphi.github.io/epidatr/dev/reference/get_api_key.html) for example).

dshemetov:
Yea, that'll be hard. I think we found the minimal effort, maximal user gain solution, so we should probably just go with it.

brookslogan:
btw, this document command looks super handy with GitHub suggestions.  Very cool.  (And doesn't seem like there should be conflicts if you need to rebase local changes made in parallel to the doc commits.)

dshemetov:
Right? It works well enough. (Wonder if we can have CI trigger from its commits too, because sometimes people push without documenting NAMESPACE changes, the CI breaks, but then the CI document pushes a change, but the status of the CI doesn't change to passing.)

PR #218, opened 2023-11-18
Add more detail to `avail_endpoint` descriptions

nmdefries:
Clarify data source, region, and what types of diseases and signals are available. Add message that most data is US only.


R/endpoints.R
```diff
@@ -1103,11 +1104,11 @@ pub_fluview_meta <- function(fetch_args = fetch_args_list()) {
       create_epidata_field_info("latest_issue", "date"),
       create_epidata_field_info("table_rows", "int")
     )
-  )
+  ) %>% fetch(fetch_args = fetch_args)
```
dshemetov, line 1107: Nice catch!

dshemetov (approved):
Looks good!

Issue #202, opened 2023-11-01
`pub_covid_hosp_state_timeseries` `issue` column, and likely related endpoints/columns, are not parsed successfully


brookslogan:
``` r
library(epidatr)
covid_hosp_timeseries_selection <-
  pub_covid_hosp_state_timeseries("ak", as.Date("2022-12-17"))
covid_hosp_timeseries_selection$issue
#> [1] NA
# VS:
response <-
  pub_covid_hosp_state_timeseries("ak", as.Date("2022-12-17"), fetch_args = fetch_args_list(dry_run = TRUE)) |>
  epidatr:::request_url() |>
  httr::GET()
httr::content(response)$epidata[[1]]$issue
#> [1] 20231027
```

<sup>Created on 2023-11-01 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

brookslogan:
Workaround: `disable_date_parsing = TRUE` in `fetch_args_list`, then convert the impacted columns with e.g.
```r
            date = as.Date(as.character(date), format="%Y%m%d"),
            issue = as.Date(as.character(issue), format = "%Y%m%d"),
```

PR #220, opened 2023-11-27
Correctly parse issue field for `covidcast_hosp_state_timeseries`

nmdefries:
Closes https://github.com/cmu-delphi/epidatr/issues/202

### DIFF
--- a/R/epidatacall.R
+++ b/R/epidatacall.R
@@ -41,12 +41,31 @@
 #' @return
 #' - For `create_epidata_call`: an `epidata_call` object
 #'
+#' @importFrom purrr map_chr
 create_epidata_call <- function(endpoint, params, meta = NULL,
                                 only_supports_classic = FALSE) {
   stopifnot(is.character(endpoint), length(endpoint) == 1)
   stopifnot(is.list(params))
   stopifnot(is.null(meta) || is.list(meta))
   stopifnot(is.logical(only_supports_classic), length(only_supports_classic) == 1)
+
+  if (length(unique(meta)) != length(meta)) {
+    cli::cli_abort(c(
+      "List of expected epidata fields contains duplicate entries",
+      "i" = "duplicates in meta can cause problems parsing fetched data",
+      "Please fix in `endpoints.R`"
+    ))
+  }
+
+  meta_field_names <- map_chr(meta, ~ .x$name)
+  if (length(meta_field_names) != length(unique(meta_field_names))) {
+    cli::cli_abort(c(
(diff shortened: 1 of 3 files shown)


R/model.R
```diff
@@ -138,13 +138,30 @@ parse_value <- function(info, value, disable_date_parsing = FALSE) {
...
   }
+
+  meta_field_names <- map_chr(meta, ~ .x$name)
+  missing_fields <- setdiff(names(df), meta_field_names)
+  if (
+    length(missing_fields) != 0
+  ) {
+    cli::cli_warn(
+      c(
+        "Not all return columns are specified as expected epidata fields",
```
nmdefries, line 158: Adding specs for these in another PR

nmdefries: https://github.com/cmu-delphi/epidatr/pull/223

dshemetov (approved):
lgtm, couple small suggestions to tighten up the type expectations in `create_epidata_call`

R/epidatacall.R
```diff
@@ -41,12 +41,37 @@
...
+      c(
+        "List of expected epidata fields contains duplicate entries",
+        "i" = "duplicates in meta can cause problems parsing fetched data",
+        "Please fix in `endpoints.R`"
+      ),
+      class = "epidatr__duplicate_meta_entries"
+    )
+  }
+
+  meta_field_names <- map_chr(meta, ~ .x$name)
```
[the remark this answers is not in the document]
nmdefries: RE checking meta inheritance, good idea. I'm going to put that in `create_epidata_call` so that we can catch it farther upstream.

dsweber2 (approved):

R/model.R
nmdefries: Most endpoint fields were already fully specified. #223 adds some that weren't included. @lcbrooks or @dshemetov any context on why we specify headers?

FWIW headers shouldn't change often/fast. Most endpoints aren't adding new fields. Many aren't being updated anymore and at least covidcast uses the same header for every signal.

dsweber2: It just worries me as yet another place where we're duplicating state. I hadn't caught that it was already pretty thoroughly covered though; in for a penny, in for a pound I guess

dshemetov: It's something we inherited from Sam's original development effort. I am guessing he ran into issues where he couldn't rely on automatic type inference from the JSON output and went in for hard coding all data types.

It's definitely a maintenance burden. We could try running an experiment where we parse the JSON output just using tibbles automatic type casting and see if we have any bad cases.

PR #223, opened 2023-11-28
fill in ? for public endpoints

nmdefries:


dshemetov (approved):
looks good, i didn't check the column names for correctness or completeness. 

btw, what happens if the metadata for an endpoint is 
- missing a field that comes back from a response? does it still get returned in the tibble or is it missing?
- has an extra field not in the response? is there an error or is it silently ignored?

Issue #194, opened 2023-10-06
`as_of` can't handle Date-type input


dshemetov:
```R
r$> forecast_due_date = as.Date("2021-10-19")

r$> epidatr::pub_covidcast(
          "hhs",
          "confirmed_admissions_influenza_1d",
          "state",
          "day",
          "*",
          epirange(forecast_due_date - 20L, forecast_due_date),
          as_of = forecast_due_date
        )
Error in charToDate(x) : 
  character string is not in a standard unambiguous format

r$> epidatr::pub_covidcast(
          "hhs",
          "confirmed_admissions_influenza_1d",
          "state",
          "day",
          "*",
          epirange(forecast_due_date - 20L, forecast_due_date),
          as_of = "2021-10-19"
        )
# A tibble: 1,080 × 15
...
```

dsweber2:
pretty sure this is a dupe of #192, he mentions length 1 vectors as well

dshemetov:
Seems right, but let's keep this open so we make sure we address the example above with whatever fix we implement

nmdefries:
Closed by https://github.com/cmu-delphi/epidatr/pull/224

Issue #192, opened 2023-10-02
Error when checking if `as_of="*"`


brookslogan:
``` r
library(epidatr)
set_cache(days = 28L, max_size = 2e3)
pub_covidcast("hhs", "confirmed_7dav_incidence_prop", "state", "day", "*", "*", as_of=Sys.Date())
#> Error in charToDate(x): character string is not in a standard unambiguous format
```

<sup>Created on 2023-10-02 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

This appears to be because of this comparison: `epidata_call$params$as_of != "*"` which appears to attempt date conversion on `"*"`.  Note we don't even support `as_of="*"` in the API, which would be extremely large (returning full snapshots for every version), only `issues="*"`.

brookslogan:
Same thing happens when specifying a single date in `issues =`, and some other error message or warning + bad behavior may happen with `issues =` date vector.

Maybe these checks should be done with `identical`.

brookslogan:
This is actually blocking use of `as_of` and `issues` when using a single `Date` object, regardless of whether we are using caching or not.

PR #224, opened 2023-11-29
Support version params `as_of` and `issues` passed as Date objects

nmdefries:
Closes [#192](https://github.com/cmu-delphi/epidatr/issues/192) and #194.

When `as_of` or `issues` is provided a Date, the comparison `as_of != "*"` tries and fails to convert "\*" to a Date also so it can do the comparison. Instead, use `identical`, which doesn't try to implicitly convert the "\*".

### DIFF
--- a/tests/testthat/test-utils.R
+++ b/tests/testthat/test-utils.R
@@ -58,7 +58,7 @@ test_that("check_is_cachable can handle both str and date inputs of various leng
   epidata_call$params$as_of <- NULL
   epidata_call$params$issues <- c(as.Date("2022-01-01"), as.Date("2022-02-01"))
   expect_no_error(check_is_cachable(epidata_call, fetch_args))
-  
+
   # issues epirange
   epidata_call$params$as_of <- NULL
   epidata_call$params$issues <- epirange(as.Date("2022-01-01"), as.Date("2022-02-01"))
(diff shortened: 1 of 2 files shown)


tests/testthat/test-utils.R
```diff
@@ -20,3 +20,42 @@ test_that("format_list", {
...
+
+  # issues single date
+  epidata_call$params$as_of <- NULL
+  epidata_call$params$issues <- as.Date("2022-01-01")
+  expect_no_error(check_is_cachable(epidata_call, fetch_args))
+
+  # issues date vector
+  epidata_call$params$as_of <- NULL
+  epidata_call$params$issues <- c(as.Date("2022-01-01"), as.Date("2022-02-01"))
+  expect_no_error(check_is_cachable(epidata_call, fetch_args))
```
dshemetov, line 60:
```suggestion
  expect_no_error(check_is_cachable(epidata_call, fetch_args))
  
  # issues epirange
  epidata_call$params$as_of <- NULL
  epidata_call$params$issues <- epirange(as.Date("2022-01-01"), as.Date("2022-02-01"))
  expect_no_error(check_is_cachable(epidata_call, fetch_args))
```

issues can also be an epirange

dshemetov (changes requested):
Looks good, added one more test case

dshemetov (approved):
🚢 🚢 🚢

Issue #212, opened 2023-11-08
Nat dogfooding notes


nmdefries:
- Documentation website
    - [x] ~the [epidatr vignette](https://cmu-delphi.github.io/epidatr/articles/epidatr.html) isn't up to date about ways to set api keys -- it doesn't mention the `set_api_key` fn.~ code is up to date, but site needs to be released (looks like releases not done automatically(
        - [x] "signal discovery" how-to should be more fleshed out and NOT listed as advanced or experimental. Potentially better as a stand-alone vignette
        - [x] Can copy over some `covidcast` documentation, for both [signal discovery](https://cmu-delphi.github.io/covidcast/covidcastR/articles/covidcast.html#finding-signals-of-interest) and other
        - [x] [geo discovery/how to find geo ids](https://cmu-delphi.github.io/covidcast/covidcastR/articles/covidcast.html#finding-counties-and-metro-areas) (basic version done)
        - [x] [how to use metadata](https://cmu-delphi.github.io/covidcast/covidcastR/articles/covidcast.html#signal-metadata)
        - [x] [versioned data](https://cmu-delphi.github.io/covidcast/covidcastR/articles/covidcast.html#tracking-issues-and-updates) (separate vignette)
        - [x] getting the call obj shouldn't be part of "getting started"
        - [x] And then it just has a huge list of endpoints?? doesn't seem super useful since it doesn't actually describe them, just link to the online docs and provide the same example from each fn doc page
    - [x] [landing page](https://cmu-delphi.github.io/epidatr/index.html) looks more like "getting started". Should focus more on what the package is/provides, how it differs from `covidcast`, setup, links to other resources, good vignettes to start with, etc
    - [ ] #238
        - [x] private endpoints should be at the bottom (or maybe even hidden/collapsed by default) since they're not relevant to most users.
- Function documentation (most of these are really issues with the [Epidata doc website](https://cmu-delphi.github.io/delphi-epidata/))
    - [x] `fetch` description typo. "only support will return"
    - [x] ~`timeset` doc page doesn't link to any of the formats it mentions. Esp looking at epiweek and epirange. And then asking for docs on `epiweek` sends me to a lubridate page. Is there even an `EpiWeek` fn/class available?~ Fixed in a previous doc update
    - [x] docs for `covidcast_epidata` refer to `covidcast` fn that no longer exists. "These objects can be used directly to fetch data, without requiring us to use the covidcast() function."
[...]

PR #230, opened 2023-12-08
Redo landing page and intro doc

nmdefries:
- Add more descriptive info to landing page
- Add citation and licensing info to landing page
- Move installation, setup, and basic usage from landing page to intro vignette
- Move signal discovery and example "other" endpoint calls from intro vignette to new signal discovery vignette
- Add geo info to intro vignette
- Add examples of geo specification to intro vignette
- Add examples of `as_of` specification to intro vignette
- Add stub versioned data vignette

Partially addresses https://github.com/cmu-delphi/epidatr/issues/212

### DIFF
--- a/vignettes/epidatr.Rmd
+++ b/vignettes/epidatr.Rmd
@@ -165,9 +165,14 @@ ggplot(epidata, aes(x = time_value, y = value)) +
 
 ## Signal discovery
 
-The [Delphi Epidata API documentation](https://cmu-delphi.github.io/delphi-epidata/) has more information on available endpoints. You can also use the `avail_endpoints()` function to get a table of endpoint functions:
+Above we used data from [Delphi’s symptom surveys](https://delphi.cmu.edu/covid19/ctis/), but the Epidata API includes numerous data streams: medical claims data, cases and deaths, mobility, and many others. This can make it a challenge to find the data stream that you are most interested in.
+
+The Epidata documentation lists all the data sources and signals available through the API for [COVID-19](https://cmu-delphi.github.io/delphi-epidata/api/covidcast_signals.html) and for [other diseases](https://cmu-delphi.github.io/delphi-epidata/api/README.html#source-specific-parameters).
+
+You can also use the `avail_endpoints()` function to get a table of endpoint functions:
 
 ```{r}
 avail_endpoints()
 ```
 
+See `vignette("signal-discovery")` for more information.
--- a/vignettes/epidatr.Rmd
+++ b/vignettes/epidatr.Rmd
@@ -148,9 +148,10 @@ knitr::kable(head(epidata))
 
 ## Getting versioned data
 
-The Epidata API stores a historical record of all data, including corrections and updates, which is particularly useful for accurately backtesting forecasting models. To fetch versioned data
-
-We can also request data for a single location at a time, via the `geo_values` argument.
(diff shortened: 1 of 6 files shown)


vignettes/epidatr.Rmd
```diff
@@ -11,379 +11,224 @@ vignette: >
...
-API docs: <https://cmu-delphi.github.io/delphi-epidata/api/ecdc_ili.html>
 
-```{r, eval = FALSE}
-pub_ecdc_ili(regions = "Armenia", epiweeks = 201840)
-```
+## Plotting
 
-#### KCDC ILI
+Because the output data is in a standard `tibble` format, we can easily plot
+it using `ggplot2`:
```
nmdefries, line 176: Should we suggest any particular plotting libraries, especially for more complex plots like choropleths?

vignettes/epidatr.Rmd
```diff
@@ -11,379 +11,224 @@ vignette: >
...
-Example queries with all the endpoint functions available in this package are given [below](#example-queries).
+To generate your key,
+[register for a pseudo-anonymous account](https://api.delphi.cmu.edu/epidata/admin/registration_form).
+See the `save_api_key()` function documentation for details on how to set up
+`epidatr` to use your API key.
 
-## Advanced Usage (Experimental)
+_Note_ that private endpoints (i.e. those prefixed with `pvt_`) require a
+separate key that needs to be passed as an argument. These endpoints require
+specific data use agreements to access.
```
nmdefries, line 57: Need to include more info here. How can users get access to private endpoints? Do we have a page or contact email for that?

dshemetov:
Are there parts you would like specific feedback on here? I've read so many variations on these docs that it's hard for me to focus and make helpful suggestions. If you don't need specific feedback, I'd be happy to just trust your taste in re-organizing this information.

vignettes/epidatr.Rmd
dshemetov: Never found that information anywhere. @brookslogan do you know?

vignettes/epidatr.Rmd
dshemetov: A choropleth example would be neat if there's a package we could use, but it's pretty low priority imo.

nmdefries:
No specific feedback desired. Give as much or as little feedback as you want.

The changes here can be summarized as:

- give contextual, not usage, info on the landing page
- add more, shorter vignettes
- add more depth and specifics on usage

dshemetov (approved):
Trusting Nat's reorganization here. If there are typos, small edits needed, we can catch them after this is merged into dev and we work on other PRs for the next version.

vignettes/epidatr.Rmd
brookslogan: I don't.  Pre-COVID I don't think there was even a procedure, but I'm not sure if there is one now.

Context: Pre-COVID I think most of these we probably didn't even have permission to re-share.  Many were for internal use (e.g., feeding into ILI-Nearby) + potentially sharing back/visualizing to the providers or maybe very limited pre-agreed-upon third parties (and even this last possibility I'm not sure ever actually happened).  One that I know of (NoroSTAT) was private just for testing purposes; it requires revival work now though so I'm not sure there's a point in broadening access.

nmdefries: Will look into this and address separately.

Issue #51, opened 2022-09-30
Add `print` S3 methods for any classes missing them (e.g., `EpiRange`)


brookslogan:
The highest-priority things to get `print` methods already have methods written, but may not be used; see #43.  However, classes like `EpiRange` may still benefit from a neater-looking `print` method.

capnrefsmmat:
Seconding that EpiRange in particular needs a print method. For daily and weekly data, they currently print like this:

```r
epirange(as.Date("2021-01-01"), as.Date("2021-01-07"))
#> $from
#> [1] 20210101
#> 
#> $to
#> [1] 20210107
#> 
#> attr(,"class")
#> [1] "EpiRange"

epirange(202202, 202204)
#> $from
#> [1] 202202
#> 
#> $to
#> [1] 202204
#> 
#> attr(,"class")
#> [1] "EpiRange"
```

I think there should be a simple human-readable format, that (a) clearly states if it's daily or weekly, (b) puts the dates in a standard format like ISO 8601 instead of mashing components together with no spacing, and (c) is nice and compact on one line.

PR #232, opened 2024-01-04
Add print methods for rest of custom classes

nmdefries:
Closes https://github.com/cmu-delphi/epidatr/issues/51


R/epidatacall.R
```diff
@@ -198,6 +198,14 @@ fetch_args_list <- function(
   )
 }
 
+#' @export
+print.fetch_args <- function(x, ...) {
+  stopifnot(inherits(x, "fetch_args"))
+  cli::cli_h1("<fetch_args> object:")
+  # Print all non-class fields.
+  print(x[attr(x, "names")])
```
brookslogan, line 206: issue?: I think this probably does both printing via messages and via regular prints to stdout.  This causes headaches when you want to build .Rmds and/or `capture.output`.  See cmu-delphi/epipredict#277.

R/epidatacall.R
```diff
@@ -198,6 +198,14 @@ fetch_args_list <- function(
   )
 }
 
+#' @export
+print.fetch_args <- function(x, ...) {
+  stopifnot(inherits(x, "fetch_args"))
```
brookslogan, line 203: musing: I think the conventional S3 method pattern is omit this type of check, and assume that you did / could have gotten here from S3 dispatch from the generic.  Same thing probably applies to other `print` implementations added below.

R/model.R
```diff
@@ -61,6 +61,31 @@ epirange <- function(from, to) {
...
+  stopifnot(inherits(x, "EpiRange"))
+
+  if (nchar(x$from) == 8) {
+    date_type <- "Days" # nolint: object_usage_linter
+    x$from <- as.Date(as.character(x$from), "%Y%m%d")
+    x$to <- as.Date(as.character(x$to), "%Y%m%d")
+  } else if (nchar(x$from) == 6) {
+    date_type <- "Epiweeks" # nolint: object_usage_linter
+    x$from <- format(
+      as.Date(as.character(x$from), "%Y%U"),
```
brookslogan, line 75: issue: `%U` != US epiweek convention.  Both use Sunday as start of the week, but in epiweek convention, weeks are numbered 1 to 53 and assigned to a year based on what year their Wednesday falls on.

issue: this always outputs week 0 for some reason.  Maybe `as.Date` failing due to a wday not being provided?
```
epirange(200101, 200230)

── <EpiRange> object: ────────────────────────────────────────────────────────────────────────
Epiweeks from 2001w00 to 2002w00
```

brookslogan (changes requested):
Nice rounding out of the print functionality.  There appears to be some bugs with epiweek printing, plus some lower-priority annoying stuff that likely happens if you try to use this in an Rmd/qmd.

R/model.R
nmdefries: Good catch! Avoid date conversion for simplicity.

R/epidatacall.R
nmdefries: Used `cli::cli_dl` to print name-value pairs. Most `cli` functions drop item names.

dshemetov (approved):

R/model.R
brookslogan: Sounds good; that's also more general.  Will be good if we add non-US-epiweeks, though the current wording ("Epiweeks") will need an update if we do.  (I don't think we have non-US-epiweeks anywhere yet (besides the `collection_weeks` which are actually specified as dates).)

brookslogan (approved):

PR #234, opened 2024-01-05
Make endpoints fetch all dates by default

nmdefries:
Partially addresses https://github.com/cmu-delphi/epidatr/issues/212.

Fetch all dates (time values) by default, using "\*". Also support user-provided "\*" with addition of helper function to convert "\*" into a long `epirange`.

Given confusing naming of date params for `pub_covid_hosp_facility`, where `collection_weeks` actually requires YYYYMMDD format, let users provide weeks (with a warning).

Note: `pub_wiki` and `pvt_twitter` aren't updated here. They both take _either_ days or epiweeks (mutually exclusive), so I'm proposing an interface change first in https://github.com/cmu-delphi/epidatr/pull/236.


dshemetov (approved):
High level skim, looks good to me. Might be good to add a changelog line now so you don't have to later.

melange396:
be careful with this; two months from now, if just 2 county-level daily signals that started January 2020 are requested simultaneously, they could exceed the row limit: `2*3295*(4*365+2*30) > 1e7`.  it looks like [the user will be notified](https://github.com/cmu-delphi/epidatr/blob/ff68c4551aaf375c7b7a5fe4621f16ae24eed2be/R/epidatacall.R#L295-L305) if the [results are truncated](https://github.com/cmu-delphi/delphi-epidata/blob/81b9ac08f617924f3b47c24c97420c29bf4cf4bd/src/server/_printer.py#L139-L140), but it could still be a waste of time and cycles.

dshemetov:
Good point. Maybe add a warning about county=* queries?

nmdefries:
Thank for the information, george. Good to know we have a truncation warning already set up.

I believe only `covidcast` supports requesting all regions via the `"*"` wildcard (I'm can't check the `pvt_` endpoints). `pvt_twitter` only [reports down to the state level](https://cmu-delphi.github.io/delphi-epidata/api/twitter.html#twitter-stream-data) as the smallest region. `pub_wiki` doesn't take regions at all, but also doesn't take a wildcard for article names. So for these two endpoints, hitting the row limit seems hard or impossible.

Ideally, these endpoints would support "*" for geos, but that will involve a lot more work. I'll keep the row limit in mind if/when we get to that!

PR #236, opened 2024-01-09
Combine timetype wiki twitter

nmdefries:
`pvt_twitter` and `pub_wiki` can take dates in either day or week format. These are provided as separate, mutually exclusive (will fail) params `dates` and `epiweeks`. This format disagrees with the interface for `pub_covidcast`, which also supports days and weeks. Change the twitter and wiki interfaces to match, with new params `time_type` and `time_values`.

This also makes supporting the date range default to wildcard "\*" (all dates) easier -- if both `dates` and `epiweeks` params were set to "\*" by default, the user would have to specify which date type they wanted to use anyway, a la `time_type`, since we wouldn't be able to automatically determine precedence.

### DIFF
--- a/R/endpoints.R
+++ b/R/endpoints.R
@@ -2094,15 +2094,17 @@ pvt_sensors <- function(
 #' pvt_twitter(
 #'   auth = Sys.getenv("SECRET_API_AUTH_TWITTER"),
 #'   locations = "CA",
-#'   epiweeks = epirange(201501, 202001)
+#'   time_type = "week",
+#'   time_values = epirange(201501, 202001)
 #' )
 #' }
 #' @param auth string. Restricted access key (not the same as API key).
 #' @param locations character. Locations to fetch.
 #' @param ... not used for values, forces later arguments to bind by name
-#' @param dates [`timeset`]. Dates to fetch. Mutually exclusive with `epiweeks`.
-#' @param epiweeks [`timeset`]. Epiweeks to fetch. Mutually exclusive with
-#' `dates`.
+#' @param time_type string. The temporal resolution of the data (either "day" or
+#'  "week", depending on signal).
+#' @param time_values [`timeset`]. Dates or epiweeks to fetch. Defaults to all
+#'  ("*") dates.
 #' @param fetch_args [`fetch_args`]. Additional arguments to pass to `fetch()`.
 #' @return [`tibble::tibble`]
 #' @keywords endpoint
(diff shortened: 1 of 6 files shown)


R/endpoints.R
```diff
@@ -2111,21 +2115,31 @@ pvt_twitter <- function(
...
     locations,
     ...,
-    dates = NULL,
-    epiweeks = NULL,
+    time_type = c("day", "week"),
+    time_values = "*",
     fetch_args = fetch_args_list()) {
   rlang::check_dots_empty()
 
+  time_type <- match.arg(time_type)
```
dsweber2, line 2123: I'm not entirely sure this is necessary, or else `pub_covidcast` is missing something important, since it doesn't do this. I would maybe make a function for this bit of logic if it is necessary, since it (should?) be included 3 times.

dsweber2 (approved):
The tests cover what I would think of. Only 1 minor note.

I'm a bit surprised that these weren't following the convention of the other endpoints. Looking for other exceptions, I didn't find any, thankfully. Good catch!
- CDC matches the api in using `epiweeks`
- `pub_covid_hosp_state_timeseries` uses `dates`, matching the API
- `pub_delphi` does actually use `epiweek`s only
- same for `pub_dengue_nowcast`, `pvt_dengue_sensors`, `pub_ecdc_ili`, `pub_flusurv`, `pub_fluview_clinical`, `pub_fluview`, `pub_gft`, `pub_kcdc_ili`, `pub_nidss_dengue`, and `pub_nidss_flu`
- I'm not actually sure about `pvt_norostat`. The docs have [ellipses](https://cmu-delphi.github.io/delphi-epidata/api/norostat.html) covering the date. 
- `pub_nowcast`, as `epiweeks` matches their api docs
- `pub_paho_dengue` has ellipses covering the date
- `pvt_quidel`, `pvt_sensors`,  use `epiweeks` in both

dshemetov (approved):

R/endpoints.R
capnrefsmmat: Here it's necessary because the default value is `c("day", "week")`, but you can only have one `time_type`. `match.arg()` takes the first if the default is not overridden. It's a common convention in R that if the argument can be one of several string values, you make the default list all of them, and `match.arg()` takes care of picking the first (if none is provided) or ensuring the provided value is valid otherwise.

`pub_covidcast()` doesn't set a default for this argument, so it's not strictly necessary. However, `pub_covidcast()` could do `match.arg(time_type, c("day", "week"))` to enforce that the type must be one or the other but not both.

PR #237, opened 2024-01-10
Parse int fields as double so large values don't get set to NA

nmdefries:
Some values (e.g. 2542264861) returned by `pub_wiki` are being coerced to NA [here](https://github.com/cmu-delphi/epidatr/blob/ff68c4551aaf375c7b7a5fe4621f16ae24eed2be/R/model.R#L159) because they exceed R's int storage capacity.

### DIFF
--- a/R/model.R
+++ b/R/model.R
@@ -157,7 +157,18 @@ parse_value <- function(info, value, disable_date_parsing = FALSE) {
     return(as.logical(value))
   } else if (info$type == "int") {
     # Int doesn't have enough capacity to store some weekly `pub_wiki` values.
-    return(as.double(value))
+    value <- as.double(value)
+    if (any(value != round(value))) {
+      cli::cli_warn(
+        c(
+          "Values in {info$name} were expected to be integers but contain a decimal component",
+          "i" = "Decimal components are returned as-is"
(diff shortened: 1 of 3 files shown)


R/model.R
```diff
@@ -156,7 +156,19 @@ parse_value <- function(info, value, disable_date_parsing = FALSE) {
...
+    if (any(value != round(value))) {
+      cli::cli_warn(
+        c(
+          "Values in {info$name} were expected to be integers but contain a decimal component",
+          "i" = "Decimal components are returned as-is"
+        ),
+        class = "epidatr__int_nonzero_decimal_digits"
+      )
+    }
+
```
dshemetov, line 170: Could use this: https://mllg.github.io/checkmate/reference/checkIntegerish.html

dshemetov (approved):
Looks good, minor non-blocking suggest

R/model.R
nmdefries: Thanks, that's cleaner. Unfortunately, their check is too thorough -- it also fails when values are too large to be stored as ints, which is the case here.

dshemetov: Oof, too bad

PR #247, opened 2024-02-16
ci+docs: match style with epipredict

dshemetov:
Header bar text is unreadable on [epidatr/dev](https://cmu-delphi.github.io/epidatr/dev/) docs, fix by matching the settings in `_pkgdown.yml` with the [settings in epipredict ](https://github.com/cmu-delphi/epipredict/blob/dev/_pkgdown.yml), which [look correct](https://cmu-delphi.github.io/epipredict/). I tested this locally with: `build_home_index(); init_site()` and then `python -m http.server` to host.

image

Also matches the pkgdown ci with how it's done in epiprocess (based on branch instead of .9000 version numbers).

### DIFF
--- a/_pkgdown.yml
+++ b/_pkgdown.yml
@@ -1,37 +1,29 @@
+# Colors should stay consistent across epipredict & epidatr, using Carnegie
+# Red https://www.cmu.edu/brand/brand-guidelines/visual-identity/colors.html
 template:
   bootstrap: 5
   bootswatch: cosmo
   bslib:
     font_scale: 1.0
-    primary: '#C41230'
-    link-color: '#C41230'
-    navbar-bg: '#C41230'
-    navbar-fg: '#f8f8f8'
-
-development:
-  mode: auto
+    primary: "#C41230"
+    link-color: "#C41230"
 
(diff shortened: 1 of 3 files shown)


_pkgdown.yml
```diff
@@ -1,37 +1,29 @@
...
-development:
-  mode: auto
+    primary: "#C41230"
+    link-color: "#C41230"
 
 navbar:
-  bg: '#C41230'
-  fg: '#f8f8f8'
+  bg: primary
+  type: dark
```
brookslogan, line 13: doesn't seem to match epiprocess [here](https://github.com/cmu-delphi/epiprocess/blob/fd2339d8379455fda1ff27cc078fb23be5c3a93c/_pkgdown.yml#L7).  Are they actually the same or do we need to update one of them?

dshemetov: I think we need to update epiprocess because its [docs don't have the right colors either](https://cmu-delphi.github.io/epiprocess/dev/) (though they don't have the strange invisibility problem).

nmdefries (approved):

PR #248, opened 2024-02-16
fix: make default cache directory R version portable

dshemetov:
Use `rappdirs::user_cache_dir("R", version = "epidatr")` instead of `tools::R_user_dir("epidatr")` ([previous discussion](https://github.com/cmu-delphi/epidatr/issues/181#issuecomment-1745458917)).

Also informs the user of the cache and its variables when its used, e.g.:

```r
! epidatr cache is being used (set env var EPIDATR_USE_CACHE=FALSE if not intended).
ℹ The cache directory is ~/.cache/R/epidatr.
ℹ The cache will be cleared after 1 days and will be pruned if it exceeds 1024 MB.
ℹ The log of cache transactions is stored at ~/.cache/R/epidatr/logfile.txt.
```

### DIFF
--- a/R/cache.R
+++ b/R/cache.R
@@ -154,7 +154,6 @@ set_cache <- function(cache_dir = NULL,
     }
   }
 
-
   if (!cache_usable) {
     print(glue::glue(
       "The directory at {cache_dir} is not accessible; check permissions and/or use a different ",
@@ -168,6 +167,13 @@ set_cache <- function(cache_dir = NULL,
       logfile = file.path(cache_dir, logfile)
     )
   }
+
+  cli::cli_inform(c(
+    "!" = "epidatr cache is being used (set env var EPIDATR_USE_CACHE=FALSE if not intended).",
+    "i" = "The cache directory is {cache_dir}.",
+    "i" = "The cache will be cleared after {days} days and will be pruned if it exceeds {max_size} MB.",
+    "i" = "The log of cache transactions is stored at {file.path(cache_dir, logfile)}."
+  ))
 }
 
 #' Manually reset the cache, deleting all currently saved data and starting afresh
--- a/R/cache.R
+++ b/R/cache.R
@@ -76,9 +76,9 @@ cache_environ$epidatr_cache <- NULL
 #' )
 #'
 #' @param cache_dir the directory in which the cache is stored. By default, this
-#'   is `tools::R_user_dir()` if on R 4.0+, but must be specified for earlier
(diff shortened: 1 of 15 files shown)


R/cache.R
```diff
@@ -104,7 +104,7 @@ set_cache <- function(cache_dir = NULL,
                       logfile = NULL,
                       confirm = TRUE) {
   if (is.null(cache_dir) && sessionInfo()$R.version$major >= 4) {
-    cache_dir <- Sys.getenv("EPIDATR_CACHE_DIR", unset = tools::R_user_dir("epidatr"))
+    cache_dir <- Sys.getenv("EPIDATR_CACHE_DIR", unset = rappdirs::user_cache_dir("R", version = "epidatr"))
```
brookslogan, line 107: this is still behind >= 4 check.  is that intended?

brookslogan: also these are different paths.  are they both valid?

brookslogan: oh right we prob should have been doing `tools::R_user_dir("epidatr", which = "cache")`

dshemetov: Ah ty!

dsweber2 (approved):
Not sure whether this is CRAN approved or not.

> Packages should not write in the user’s home filespace (including clipboards), nor anywhere else on the file system apart from the R session’s temporary directory (or during installation in the location pointed to by TMPDIR: and such usage should be cleaned up). Installing into the system’s R installation (e.g., scripts to its bin directory) is not allowed.

> Limited exceptions may be allowed in interactive sessions if the package obtains confirmation from the user.

> For R version 4.0 or later (hence a version dependency is required or only conditional use is possible), packages may store user-specific data, configuration and cache files in their respective user directories obtained from tools::R_user_dir(), provided that by default sizes are kept as small as possible and the contents are actively managed (including removing outdated material). 

`rappdirs::user_cache_dir` isn't mentioned here, so it may or may not generate some flak. It does generate the same path on my ubuntu machine though, so it's probably fine

dshemetov:
I think it's worth possibly bumping against the CRAN policy here just to understand that paragraph you linked better. A strict interpretation doesn't make sense, since that would mean any sort of function that writes a file to a user-specified path is not allowed, and that's not the case. I don't see how someone calling `set_cache` to write files is that much different. Gabor Csardi (r-lib maintainer) [pointed out on the mailing list in 2018](https://www.mail-archive.com/r-package-devel@r-project.org/msg02460.html) that this policy goes against industry standards.

PR #249, opened 2024-02-16
doc: update readme with blog post feedback

dshemetov:
Incorporating feedback from [here](https://github.com/cmu-delphi/www-main/pull/912).


dsweber2 (approved):
Reads well; I'll assume it matches whatever feedback was given. It doesn't have typos or awkward phrasing as far as I can tell

PR #254, opened 2024-02-27
Bring CRAN up to date

dsweber2:
See News.md I guess


dsweber2:
one problem with submitting `dev` to CRAN is that we effectively can't add any commits to dev until CRAN has accepted. Or at least, we need to be doing a PR from a specific commit...

PR #268, opened 2024-03-04
Turn off endpoint missing checks for `geo_values` and `time_values`, which have defaults

nmdefries:
### Checklist

Please:

      PR).
      number (the second number).

`missing` doesn't count default values as non-missing. If a user doesn't pass `geo_values` or `time_values` (both of which default to "*" in `pub_covidcast`), or `dates` (in `pub_covid_hosp_state_timeseries`), the `missing` check fails. To avoid this, just don't check missingness of those two arguments.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: epidatr
 Type: Package
 Title: Client for Delphi's 'Epidata' API
-Version: 1.1.1
+Version: 1.1.3
 Authors@R:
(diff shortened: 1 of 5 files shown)


dshemetov (changes requested):
Good catch. Could you also add lines to the documentation that says that `geo_value=*` and `time_value=*` is default?

dshemetov (approved):
lgtm

PR #273, opened 2024-03-07
fix: improve reading of `EPIDATR_USE_CACHE` variable

dshemetov:
### Checklist

Please:

      PR).
      number (the second number).

`EPIDATR_USE_CACHE` only supported exactly "TRUE" before. Now it supports all logical values and includes a warning when any value that can't be converted to logical is provided.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,11 @@
+# epidatr 1.2.0
+
+## Changes
+
+- Improve handling of the `EPIDATR_USE_CACHE` environment variable, allowing it
+  to be any value convertable by `as.logical()` and handle the case when it
+  can't be converted.
+
 # epidatr 1.1.1
 
 ## Changes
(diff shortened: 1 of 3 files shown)


NEWS.md
```diff
@@ -1,10 +1,19 @@
+# epidatr 1.2.0
```
dsweber2, line 1: is this 1.2.0 instead of 1.1.3 so that when we do a release all the changes are gathered together?

dshemetov: Exactly. Figured the 1.1.1 deviation is fine, since that was a little hotfix release.

NEWS.md
```diff
@@ -1,10 +1,19 @@
...
+
+## Changes
+
+- Improve handling of the `EPIDATR_USE_CACHE` environment variable, allowing it
+  to be any value convertable by `as.logical()` and handle the case when it
+  can't be converted.
```
nmdefries, line 7: nit: please include PR number here

```suggestion
  can't be converted (#273).
```

nmdefries (approved):
Looks good

NEWS.md
dshemetov: ah my bad, I set it up to auto-merge, let's add this in a next PR

PR #301, opened 2025-02-05
feat: add ignore_cache to fetch_args_list

### LINKED ISSUE #200: Refactor idea: the `fetch` call chain is too complex
Reading through `fetch`, the callflow is:
- `fetch`
- `cache_epidata_call`
- either `fetch_tbl` or `fetch_classic`
  - `fetch_tbl` checks if the method only supports classic and errors out, but then confusingly proceeds to call `fetch_classic` anyway
  - `fetch_classic` makes a request

My preference would be:
- absorb the `fetch_tbl` and `fetch_classic` logic into a unified imperative `fetch`
- pull the actual fetching part out of `cache_epidata_call` and only do the `is_cacheable` checking and storing there after the fetch has gone through

dshemetov:
### Checklist

Please:

      PR).
      number (the second number).

- Adds the `ignore_cache` argument to `fetch_args_list()` to make it easier to temporarily ignore the cache
- Adds some documentation to the `is_cachable` property getter for clarity
- Refactors a few of the internals of fetch functions
  - Disassemble `cache_epidata_call` into pieces, move most of the pieces over to `fetch`
  - Remove `fetch_tbl` and simplify `fetch` logic
  - Consolidate cache utility functions in cache.R (check_is_cachable moved from utils.R)

- Resolves #200


brookslogan:
I'm getting some tests failing when running locally where `check_is_cachable` gives unexpected results.  There was also a CHECK warning above from not having `@importFrom tibble tibble` that seems like it still applies but might not be caught in tests.

R/epidatacall.R
```diff
@@ -228,48 +234,57 @@ fetch <- function(epidata_call, fetch_args = fetch_args_list()) {
...
-  response_content <- fetch_classic(epidata_call, fetch_args = fetch_args)
-  if (fetch_args$return_empty && length(response_content) == 0) {
-    return(tibble())
+  # Need to actually get the data, since its either not in the cache or we're not caching
+  runtime <- system.time(if (epidata_call$only_supports_classic) {
+    fetched <- fetch_classic(epidata_call, fetch_args)
+  } else {
+    response_content <- fetch_classic(epidata_call, fetch_args = fetch_args)
+    if (fetch_args$return_empty && length(response_content) == 0) {
+      fetched <- tibble()
```
brookslogan, line 276: suggest: test this branch?  CHECK was suggesting it wouldn't run due to missing importFrom.

tests/testthat/test-epidatacall.R
```diff
@@ -142,14 +144,14 @@ test_that("fetch_tbl warns on non-success", {
...
     regexp = paste0("epidata warning: `", artificial_warning, "`"),
     fixed = TRUE
   )
 })
 
-test_that("classic only fetch", {
-  # delphi is an example endpoint that only suports the classic call
+test_that("fetch classic works", {
+  # only_supports_classic is TRUE
   epidata_call <- pub_delphi(
```
brookslogan, line 155: I don't /think/ refactor changed this, but this outputs a very weird format.  Still trying to be shoehorned into a wide data.frame rather than kept as a nested list structure.

brookslogan (commented):
I have some minor suggestions that probably don't relate to the refactor itself but also include some similar cleanup, but I did not spot why my local tests might be failing; looked good to me but seems that I missed something (hence just commenting).  Here's the output if helpful:
```
ℹ Testing epidatr
! epidatr cache is being used (set env var EPIDATR_USE_CACHE=FALSE if not intended).
ℹ The cache directory is ~/.cache/R/epidatr.
ℹ The cache will be cleared after 14 days and will be pruned if it exceeds 4096 MB.
ℹ The log of cache transactions is stored at ~/.cache/R/epidatr/logfile.txt.
✔ | F W  S  OK | Context
✔ |          1 | auth
✖ | 1       42 | cache
──────────────────────────────────────────────────────────────────────────────────────────────
Failure (test-cache.R:128:7): check_is_cachable
check_is_cachable(epidata_call, fetch_args) is not FALSE

`actual`:   TRUE 
`expected`: FALSE
Backtrace:
    ▆
 1. └─epidatr (local) check_fun(...) at test-cache.R:142:3
 2.   └─testthat::expect_false(check_is_cachable(epidata_call, fetch_args)) at test-cache.R:128:7
──────────────────────────────────────────────────────────────────────────────────────────────
✔ |         19 | check
✔ |          3 | covidcast [1.2s]
✔ |      1  74 | endpoints
✔ |         12 | epidatacall
✔ |          6 | epirange
✔ |         42 | model
✔ |      1   0 | request
✔ |      1  28 | utils

══ Results ═══════════════════════════════════════════════════════════════════════════════════
Duration: 3.2 s

── Skipped tests (3) ─────────────────────────────────────────────────────────────────────────
• empty test (2): test-endpoints.R:1:1, test-utils.R:24:1
• This site is down. (1): test-request.R:2:3

── Failed tests ──────────────────────────────────────────────────────────────────────────────
Failure (test-cache.R:128:7): check_is_cachable
check_is_cachable(epidata_call, fetch_args) is not FALSE

`actual`:   TRUE 
`expected`: FALSE
Backtrace:
    ▆
 1. └─epidatr (local) check_fun(...) at test-cache.R:142:3
 2.   └─testthat::expect_false(check_is_cachable(epidata_call, fetch_args)) at test-cache.R:128:7

[ FAIL 1 | WARN 0 | SKIP 3 | PASS 227 ]
```

R/epidatacall.R
```diff
@@ -228,48 +234,57 @@ fetch <- function(epidata_call, fetch_args = fetch_args_list()) {
...
-      ),
-      epidata_call = epidata_call,
-      class = "only_supports_classic_format"
-    )
+    # Check if the data is in the cache
+    target <- request_url(epidata_call)
+    hashed <- md5(target)
+    cached <- cache_environ$epidatr_cache$get(hashed)
+    if (!is.key_missing(cached)) {
+      cli::cli_warn(
```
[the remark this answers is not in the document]
dshemetov: Yea, seems redundant, I'll remove it.

dshemetov:
Fixed a bunch of issues:
- test caught a bug (if fields is not null, then we don't cache; lost that check in the refactor), fixed that
- rewrote a few other tests
- improved tests so they actually clear and disable the cache after every test using withr::defer
- fixed incorrect parsing of only_supports_classic endpoints
- addressed other minor changes

Thanks for the review!

brookslogan:
Sorry, somehow thought this was already merged.  Taking a look now.  Quick first not: I think running tests deleted my actual cache instead of the test cache somehow?

tests/testthat/test-cache.R
```diff
@@ -18,7 +19,11 @@ test_set_cache <- function(cache_dir = new_temp_dir,
 }
 
 test_that("cache set as expected", {
+  # Setup a new cache
   expect_message(test_set_cache())
+  # Delete cache files after the test
+  withr::defer(clear_cache(disable = TRUE))
```
brookslogan, line 25: Not sure how, but one of these clear_cache() calls seems to have cleared my real cache rather than the test cache when I ran `devtools::test()`.

brookslogan (approved):
Nice catches.  I've noted a couple minor issues + request (maybe in some separate Issue/PR) to improve messaging when a cache has been cleared unexpectedly and we still try to use it.

tests/testthat/test-cache.R
dshemetov: Hm, thanks for reminding me, this happened to me too... Current guess: if you have a cache activated before tests starts, subsequent set_cache calls just reuse that cache, so then a later clear_cache destroys it. Will check.

dshemetov:
Also, I'm now thinking that `ignore_cache` is kinda weird... what doesn't seem natural is that if there is a cache and we ignore_cache, then we never update the cache. What's probably better is something like `refresh_cache`, which will fetch from API and overwrite what's in the cache currently, if the cache is on.

tests/testthat/test-cache.R
dshemetov: Alright, figured this out.

PR #330, opened 2026-01-09
fix: allows `parse_api_date` to handle various date formats and intro…

### LINKED ISSUE #245: Missing description of the `last_update` field
Problem: Notice it's missing a description of the `last_update` field. Also `last_update` # is in the wrong format. It should be a date, but it's numeric. Need to file an issue # either in `epidatr` or in `delphi-epidata`, plus update the documentation page.

Fix: meta$last_update <- as.POSIXct(meta$last_update, origin = "1970-01-01")

### LINKED ISSUE #319: bug: default `pub_covid_hosp_facility()` fails
Using the facility in the documentation example. Other facilities also fail. Note that this is the CRAN version with R 4.5.0 on a Mac.

``` r
library(epidatr)
sessioninfo::session_info()
#> ─ Session info ───────────────────────────────────────────────────────────────
#>  setting  value
#>  version  R version 4.5.0 (2025-04-11)
#>  os       macOS Sequoia 15.5
#>  system   aarch64, darwin20
#>  ui       X11
#>  language (EN)
#>  collate  en_US.UTF-8
#>  ctype    en_US.UTF-8
#>  tz       America/Vancouver
#>  date     2025-05-30
#>  pandoc   3.4 @ /Applications/RStudio.app/Contents/Resources/app/quarto/bin/tools/aarch64/ (via rmarkdown)
#>  quarto   1.6.37 @ /usr/local/bin/quarto
#> 
#> ─ Packages ───────────────────────────────────────────────────────────────────
#>  package     * version date (UTC) lib source
#>  askpass       1.2.1   2024-10-04 [1] CRAN (R 4.5.0)
#>  backports     1.5.0   2024-05-23 [1] CRAN (R 4.5.0)
#>  cachem        1.1.0   2024-05-16 [1] CRAN (R 4.5.0)
#>  checkmate     2.3.2   2024-07-29 [1] CRAN (R 4.5.0)
#>  cli           3.6.5   2025-04-23 [1] CRAN (R 4.5.0)
#>  digest        0.6.37  2024-08-19 [1] CRAN (R 4.5.0)
#>  epidatr     * 1.2.1   2025-03-20 [1] CRAN (R 4.5.0)
#>  evaluate      1.0.3   2025-01-10 [1] CRAN (R 4.5.0)
#>  fastmap       1.2.0   2024-05-15 [1] CRAN (R 4.5.0)
#>  fs            1.6.6   2025-04-12 [1] CRAN (R 4.5.0)
#>  glue          1.8.0   2024-09-30 [1] CRAN (R 4.5.0)
#>  htmltools     0.5.8.1 2024-04-04 [1] CRAN (R 4.5.0)
#>  httr          1.4.7   2023-08-15 [1] CRAN (R 4.5.0)
#>  jsonlite      2.0.0   2025-03-27 [1] CRAN (R 4.5.0)
#>  knitr         1.50    2025-03-16 [1] CRAN (R 4.5.0)
#>  lifecycle     1.0.4   2023-11-07 [1] CRAN (R 4.5.0)
#>  magrittr      2.0.3   2022-03-30 [1] CRAN (R 4.5.0)
#>  MMWRweek      0.1.3   2020-04-22 [1] CRAN (R 4.5.0)
#>  openssl       2.3.2   2025-02-03 [1] CRAN (R 4.5.0)
#>  pillar        1.10.2  2025-04-05 [1] CRAN (R 4.5.0)
#>  pkgconfig     2.0.3   2019-09-22 [1] CRAN (R 4.5.0)
#>  purrr         1.0.4   2025-02-05 [1] CRAN (R 4.5.0)
#>  R6            2.6.1   2025-02-15 [1] CRAN (R 4.5.0)
#>  reprex        2.1.1   2024-07-06 [1] CRAN (R 4.5.0)
#>  rlang         1.1.6   2025-04-11 [1] CRAN (R 4.5.0)
#>  rmarkdown     2.29    2024-11-04 [1] CRAN (R 4.5.0)
#>  rstudioapi    0.17.1  2024-10-22 [1] CRAN (R 4.5.0)
#>  sessioninfo   1.2.3   2025-02-05 [1] CRAN (R 4.5.0)
#>  tibble        3.2.1   2023-03-20 [1] CRAN (R 4.5.0)
#>  vctrs         0.6.5   2023-12-01 [1] CRAN (R 4.5.0)
#>  withr         3.0.2   2024-10-28 [1] CRAN (R 4.5.0)
#>  xfun          0.52    2025-04-02 [1] CRAN (R 4.5.0)
#>  xml2          1.3.8   2025-03-14 [1] CRAN (R 4.5.0)
#>  yaml          2.3.10  2024-07-26 [1] CRAN (R 4.5.0)
#> 
#>  [1] /Library/Frameworks/R.framework/Versions/4.5-arm64/Resources/library
#>  * ── Packages attached to the search path.
#> 
#> ──────────────────────────────────────────────────────────────────────────────
h <- pub_covid_hosp_facility("050063", epirange(20240101, 20240301)) # works
h <- pub_covid_hosp_facility("050063") # fails
#> Error in charToDate(x): character string is not in a standard unambiguous format
#> Timing stopped at: 0.136 0.005 0.601
```

<sup>Created on 2025-05-30 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

JavierMtzRdz:
### Checklist

Please:

      PR).
      number (the second number).

Sometimes, dates and timestamps are not parsed correctly because `parse_api_*` functions do not support the format in the response. For example, `pub_covid_hosp_facility("050063", fetch_args = fetch_args_list(disable_date_parsing = T))` returns dates in various formats in the same column, as shown below.
Screenshot 2026-01-08 at 4 16 32 PM
This can lead to NA results or errors, as seen in #319. Similarly, the format for the `last_update` column in `pub_covidcast_meta()`  is a timestamp, as can be [confirmed here](https://github.com/cmu-delphi/delphi-epidata/blob/2ead81fe006faaa4c4c58c309841f4a82d996aa8/src/acquisition/covidcast/database.py#L475C8-L475C57), which is unsupported.   Therefore, this PR adds `parse_api_datetime` to handle timestamp parsing and enables `parse_api_date` to manage different date formats within the same column. 

- Resolves #245, #319


R/model.R
```diff
@@ -251,7 +254,22 @@ date_to_epiweek <- function(value) {
...
+  res <- as.Date(rep(NA, length(value_char)))
+
+  for (fmt in formats) {
+    is_missing <- is.na(res) & !is.na(value_char)
+    if (!any(is_missing)) break
+    attempt <- suppressWarnings(as.Date(value_char[is_missing], format = fmt))
+    res[is_missing] <- attempt
+  }
+  res
+}
```
dshemetov, line 292: What's the advantage of writing this out versus this?
```
  as.Date(as.character(value), tryFormats = c("%Y%m%d", "%Y-%m-%d", "%m%d%Y"))
```

dshemetov (commented):
Could you add some tests of the new functions to test-model.R?

R/model.R
JavierMtzRdz: `as.Date(as.character(value), tryFormats = c("%Y%m%d", "%Y-%m-%d", "%m%d%Y"))` assumes that all elements in `value` share the same format, which can lead to `NA` or errors, such as with `pub_covid_hosp_facility("050063")` when parsing variables like `collection_weeks`. In contrast, the new version does not assume that all elements in `value` share the same format.

dshemetov: Gotcha, thanks. Could you add that as a comment on this function? This is definitely a weirdness of R's API that we're likely to forget after 3 months.

JavierMtzRdz:
The tests and comments have been included!

R/model.R
```diff
@@ -251,7 +254,25 @@ date_to_epiweek <- function(value) {
...
+
+#' @keywords internal
+parse_api_datetime <- function(value) {
+  as.POSIXct(as.numeric(value), origin = "1970-01-01")
```
dshemetov, line 296: So right now this only handles timestamps, but in 3 months, someone might use parse_api_datetime on "YYYYMMDDTHHMMSS" and forget that it's not built for that. Can we either
- rename to something like parse_timestamp_to_datetime
- or make it robust to both timestamp and that datetime format

JavierMtzRdz: I wrote a version that handles both timestamp and datetime formats. For the datetime format, I relied on an approach similar to `parse_api_date`. Given the different possible combinations, I'm wondering if it might be useful to include [`lubridate::parse_date_time`](https://lubridate.tidyverse.org/reference/parse_date_time.html) instead.

JavierMtzRdz: I guess the formats should stay fairly consistent.

dshemetov: Lubridate might be a good direction to go. I think we were hesitant about it due to GPL licensing and we didn't want to constrain our collaborators (does that sound right to you @brookslogan ?). But it looks like [lubridate has been relicensed as MIT](https://github.com/tidyverse/lubridate/issues/968#issuecomment-3657838707) (about 3 weeks ago), which removes that blocker.

JavierMtzRdz: Hi, @dshemetov! @brookslogan noted that lubridate can also influence the behaviour of other functions. Therefore, I believe we can keep the current implementation and incorporate checks for additional formats in case the API expands.

dshemetov (changes requested):
Sorry, I think `parse_api_datetime` is going to be more work to maintain than is worth. I appreciate you attempting to add the generalization, but it doesn't address timezones and handling that properly will create even more maintenance surface (i.e. opportunities for bugs).

Let's roll back to making this function specific `parse_api_timestamp_to_datetime` and just go with that. We can leave it as a research question whether lubridate could just offload all this work from us in the future.

R/endpoints.R
```diff
@@ -955,7 +955,7 @@ pub_covidcast_meta <- function(fetch_args = fetch_args_list()) {
       create_epidata_field_info("max_value", "float"),
       create_epidata_field_info("mean_value", "float"),
       create_epidata_field_info("stdev_value", "float"),
-      create_epidata_field_info("last_update", "int"),
+      create_epidata_field_info("last_update", "datetime"),
```
dshemetov: This should clear up confusion. There's some ambiguity as to whether these field_infos are the type that we get from the API or the type that we should parse into from the API. Because this metadata is mainly used by the `parse_value` function, I think the appropriate interpretation is that it is the type received from the API, so `timestamp` is the correct annotation here.
```suggestion
      create_epidata_field_info("last_update", "timestamp"),
```

R/model.R
```diff
@@ -139,6 +139,7 @@ create_epidata_field_info <- function(name,
     "int",
     "float",
     "date",
+    "datetime",
```
dshemetov:
```suggestion
    "timestamp",
```

R/model.R
```diff
@@ -170,6 +171,8 @@ parse_value <- function(info, value, disable_date_parsing = FALSE, reference_wee
     return(value)
   } else if (info$type == "date" && !disable_date_parsing && !inherits(value, "Date")) {
     return(parse_api_date(value))
+  } else if (info$type == "datetime" && !disable_date_parsing && !inherits(value, "POSIXt")) {
+    return(parse_api_datetime(value))
```
dshemetov:
```suggestion
  } else if (info$type == "timestamp" && !disable_date_parsing && !inherits(value, "POSIXt")) {
    return(parse_api_timestamp_to_datetime(value))
```

R/model.R
dshemetov, line 141:
```suggestion
#' Add metadata to an epidata_field
#' 
#' This function annotates the type of the returned API field. This is used
#' by `parse_value` downstream to determine how to convert the returned data.
create_epidata_field_info <- function(name,
```

dshemetov, line 187:
```suggestion
#' Parse data returned by the API
#' 
#' This function uses the metadata annotations to determine how to parse the incoming data.
#' The data annotation refers to the received API type, rather than the target R type.
#' @importFrom stats na.omit
```

dshemetov (changes requested):
There's ambiguity as to whether `create_epidata_field_info` annotates the received API type or the type we parse into. Following through the code, the only place that uses this metadata field is `parse_data_frame`, which in turn uses `parse_value`. It seems that this function could be interpreted in both ways. Let's just assert that it annotates the returned API type for now. We can address this differently if we choose later.

dshemetov (approved):
Minor lint bork, otherwise lgtm

JavierMtzRdz:
I updated the documentation parameters to prevent issues with the checks.

dshemetov:
Looks good

PR #333, opened 2026-01-24
Migrate HTTP requests from `httr` to `httr2`

### LINKED ISSUE #332: Migrate to `httr2`?
I've been doing some nowcasting work using `epidatr` as the API for constructing a reporting triangle e.g. https://github.com/CDCgov/nowcastNHSN/pull/12

Doing the vignette build in pkgdown I noticed that `epidatr` is still using `httr`. Its recommended to move to `httr2` here https://httr.r-lib.org/ , so I'm just mentioning that here!

JavierMtzRdz:
### Checklist

Please:

      PR).
- [X] Request a review from one of the current epidatr main reviewers:
      number (the second number).

### Migrate HTTP requests from `httr` to `httr2`

This PR rewrites `do_request`, `request_impl`, `covidcast_epidata`, and `with_base_url` to use `httr2`. The `epidata_call` S3 object was refactored to rely directly on `httr2_request`, allowing the removal of redundant fields such as `endpoint` and `params`. `base_url` is kept because the `httr2` `request` object doesn't retain it (it only classifies the URL as a hostname and a path, but doesn't distinguish between the path belonging to the base URL and the endpoint).

`R/constants.R` was also updated to define headers as a standard list, as `req_headers` receives lists.

The tests were updated to handle `httr2`. It was necessary to add `tests/testthat/helper-mocks.R` to bridge legacy test data. An alternative would be to rewrite the test data objects to generate response objects in `httr2` format. 

- Resolves #332


dshemetov (approved):
Looks pretty good to me. I needed to familiarize myself with the code, so I dove in and made some tweaks [here](https://github.com/cmu-delphi/epidatr/pull/335). If those changes look good to you, we should be good to merge both in.

JavierMtzRdz:
I'll merge your changes and remove the `httr` function from the mocks to remove `httr` from Suggests, as proposed in the tweaks PR.

JavierMtzRdz:
I realized that #336 removes the `httr` call [here](https://github.com/cmu-delphi/epidatr/pull/336/changes#diff-44dac3b5e535a50aa9f28168547c42374d8d7b9550c22905fb98c7aa433a01e2L38), so I'll keep `httr` in the suggests for this PR and remove it in #336 to avoid conflicts.

PR #334, opened 2026-02-06
Links docs

JavierMtzRdz:
### Checklist

Please:

      PR).
- [X] Request a review from one of the current epidatr main reviewers:
      number (the second number).
- [ ] Describe changes made in NEWS.md, making sure breaking changes

This PR introduces direct links to the valid geographic values, queries, sensor names, and forecasting systems detailed in the latest delphi-epidata documentation update.

### DIFF
--- a/R/endpoints.R
+++ b/R/endpoints.R
@@ -63,6 +63,9 @@ NULL
 #' )
 #'
 #' @inheritParams .epidatr_shared_params
+#' @param locations character. List of locations to fetch.
+#'   See [US Regions and States codes](https://cmu-delphi.github.io/delphi-epidata/api/geographic_codes.html#us-regions-and-states)
+#'   for details.
 #' @return [`tibble::tibble`]
 #'
 #' @keywords endpoint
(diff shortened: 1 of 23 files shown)


R/endpoints.R
```diff
@@ -63,6 +63,9 @@ NULL
 #' )
 #'
 #' @inheritParams .epidatr_shared_params
+#' @param locations character. List of locations to fetch.
+#'   See [US Regions and States codes](https://cmu-delphi.github.io/delphi-epidata/api/geographic_codes.html#us-regions-and-states) # nolint
+#'   for details.
```
dshemetov, line 68: Noticed an issue in the documentation on [this link](https://cmu-delphi.github.io/delphi-epidata/api/geographic_codes.html#national-code): sometimes US national is "us" or "usa", it depends on the endpoint.

dshemetov (approved):
lgtm

R/endpoints.R
JavierMtzRdz: As far as I remember, it happens with cases that return international-level information, such as PAHO Dengue. For those cases, I have included a reference on the corresponding endpoint page to the table of valid names. I'll do a quick check to see whether that is not the case for some of the endpoints with US data. If that is the case, I'll include the relevant changes on https://github.com/cmu-delphi/delphi-epidata/pull/1713

JavierMtzRdz: I just checked all of them, and you were right, `ght` uses "us". So, I'll add a note clarifying this case and the cases with international data.

dshemetov: Thank you!

PR #335, opened 2026-02-09
enh: httr2 tweaks

dshemetov:
### Checklist

Please:

      PR).
      number (the second number).

Some tweaks to the new httr2 approach.

- Resolves #{issue number}

### DIFF
--- a/R/covidcast.R
+++ b/R/covidcast.R
@@ -140,26 +140,9 @@ print.covidcast_data_source <- function(x, ...) {
 #' @return An instance of `covidcast_epidata`
 #' @export
 covidcast_epidata <- function(base_url = global_base_url, timeout_seconds = 30) {
-  request <- httr2::request(base_url) %>%
-    httr2::req_url_path_append("covidcast/meta") %>%
-    httr2::req_timeout(timeout_seconds)
-
-  response <- request %>%
-    httr2::req_error(is_error = function(resp) FALSE) %>%
-    httr2::req_perform()
-
-  if (httr2::resp_is_error(response)) {
-    # 500, 429, 401 are possible
-    msg <- "fetch data from API"
-    if (identical(httr2::resp_content_type(response), "text/html") && httr2::resp_has_body(response)) {
-      # grab the error information out of the returned HTML document
-      msg <- paste(msg, ":", xml2::xml_text(xml2::xml_find_all(
-        xml2::read_html(httr2::resp_body_string(response)),
-        "//p"
-      )))
-    }
-    httr2::resp_check_status(response, info = msg)
-  }
+  # covidcast_meta and covidcast/meta are two different endpoints...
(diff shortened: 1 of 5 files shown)


R/covidcast.R
```diff
@@ -140,26 +140,9 @@ print.covidcast_data_source <- function(x, ...) {
...
-      # grab the error information out of the returned HTML document
-      msg <- paste(msg, ":", xml2::xml_text(xml2::xml_find_all(
-        xml2::read_html(httr2::resp_body_string(response)),
-        "//p"
-      )))
-    }
-    httr2::resp_check_status(response, info = msg)
-  }
+  # covidcast_meta and covidcast/meta are two different endpoints...
+  response <- create_epidata_call("covidcast/meta", list()) %>%
```
JavierMtzRdz, line 144: praise: this is much simpler and efficient

R/epidatacall.R
```diff
@@ -387,32 +387,26 @@ request_url <- function(epidata_call, format_type = "classic", fields = NULL) {
...
+  } else {
+    new_scheme <- "https"
+    new_hostname <- out[[1]][1]
+  }
 
-  epidata_call$base_url <- base_url
-  epidata_call
+  epidata_call$request$url %>%
+    httr2::url_modify(scheme = new_scheme, hostname=new_hostname) %>%
+    httr2::url_parse()
```
JavierMtzRdz: I think we need an alternative for three reasons:
1. This approach does not account for the possibility that the base URL may include a path ("epidata/" in the case of the `global_base_url`).
2. We can avoid importing `stringr` since it was not previously in the package.
3. We need an `epidata_call` object.

Since my original proposal added unnecessary steps, here is a simpler approach. I also noticed we had not tested `with_base_url`, so I added some simple tests in `test-epidatacall`. This causes the checks to fail because the current version of `with_base_url` does not return an `epidata_call` object.

```suggestion
#' @keywords internal
with_base_url <- function(epidata_call, base_url) {
  stopifnot(inherits(epidata_call, "epidata_call"))
  stopifnot(is.character(base_url), length(base_url) == 1)

  old_base_url <- epidata_call$base_url
  if (endsWith(old_base_url, "/") && !endsWith(base_url, "/")) {
    base_url <- paste0(base_url, "/")
  }

  epidata_call$request$url <- sub(
    old_base_url,
    base_url,
    epidata_call$request$url,
    fixed = TRUE
  )
  epidata_call$base_url <- base_url

  epidata_call
```

JavierMtzRdz (approved):
I only have a comment regarding `with_base_url`. If you agree with the suggested approach, let me know!

R/epidatacall.R
dshemetov: Ah thanks for catching that, I didn't anticipate case (1). I like your new implementation.

JavierMtzRdz: Excellent, then I'll merge this PR.

PR #336, opened 2026-02-11
refactor: reorganize http request functions, rename core data fetching function, reorganize locations of other functions, remove unused arguments

dshemetov:
### Checklist

Please:

      PR).
      number (the second number).
- [ ] Describe changes made in NEWS.md, making sure breaking changes

A lot of long overdue simplifications in anticipation of updating this client for the new API. Will describe individual changes more thoroughly later.

Main goals:
- combine do_request and request_impl and the error checking, so the HTTP layer is encapsulated by one function
- collect user input validation functions in one file
- collect data output validation functions in one file
- clarify the dependence between fetch and fetch_classic
- rename fetch_classic to request_epidata, since classic is not very clear and, as it stands, fetch uses fetch_classic anyway
- removed format_type, debug from fetch_args_list (both unused externally)
- removed fetch_debug (just a light wrapper over request_epidata)

- Resolves #{issue number}


R/epidatacall.R
```diff
@@ -297,71 +254,49 @@ fetch <- function(epidata_call, fetch_args = fetch_args_list()) {
         response_content,
         fetch_args$disable_date_parsing,
         fetch_args$reference_week_day
-      ) %>% as_tibble()
+      ) |> as_tibble()
```
JavierMtzRdz: I'd suggest keeping magrittr's since we require R (>= 3.5.0) and the base pipe was introduced in R 4.1.0, unless we want to increase the required version. 

```suggestion
      ) %>% as_tibble()
```

R/request.R
```diff
@@ -1,18 +1,41 @@
...
 #' @importFrom httr2 req_perform req_timeout req_headers req_user_agent req_retry
 #' @importFrom httr2 req_error req_auth_basic resp_status req_method
 #' @importFrom httr2 req_body_form req_url req_url_query
+#' @importFrom xml2 read_html xml_find_all xml_text
 #' @keywords internal
-do_request <- function(epidata_call, timeout_seconds, http_method = c("GET", "POST")) {
+do_request <- function(epidata_call, format_type, timeout_seconds, fields,
+                       http_method = c("GET", "POST")) {
+  stopifnot(inherits(epidata_call, "epidata_call"))
+  stopifnot(format_type %in% c("json", "csv", "classic"))
```
JavierMtzRdz: I think we can follow the `http_method` approach to have more informative error messages. 
```suggestion
do_request <- function(epidata_call, format_type = c("json", "csv", "classic"), timeout_seconds, fields,
                       http_method = c("GET", "POST")) {
  stopifnot(inherits(epidata_call, "epidata_call"))
  format_type <- rlang::arg_match(format_type)
```

R/endpoints.R
```diff
@@ -1761,7 +1744,7 @@ pub_kcdc_ili <- function(
 #'
 #' @examples
 #' \dontrun{
-#' pvt_meta_norostat(auth = Sys.getenv("SECRET_API_AUTH_NOROSTAT"))
+#' pvt_meta_norostat(auth = Sys.getenv("DELPHI_EPIDATA_KEY"),)
```
JavierMtzRdz:
```suggestion
#' pvt_meta_norostat(auth = Sys.getenv("DELPHI_EPIDATA_KEY"))
```

R/epidatacall.R
```diff
@@ -297,71 +254,49 @@ fetch <- function(epidata_call, fetch_args = fetch_args_list()) {
...
     }
   })
 
-  # Add it to the cache if appropriate
-  if (is_cachable || (fetch_args$refresh_cache && is_cache_enabled())) {
+  # Add to cache if appropriate.
+  if (is_cachable || fetch_args$refresh_cache) {
```
JavierMtzRdz: I'd recommend going back to the previous version, as it could have problems in cases where the cache is disabled but `refresh_cache = TRUE`, which errors as in the example below. 

```r
pub_covidcast(
     source = "jhu-csse",
     signals = "confirmed_7dav_incidence_prop",
     time_type = "day",
     geo_type = "state",
     time_values = epirange(20200601, 20200801),
     geo_values = "ca",
     fetch_args = fetch_args_list(refresh_cache = TRUE)
 )
#> Error in fetch(., fetch_args = fetch_args) : 
#> attempt to apply non-function
```
```suggestion
  if (is_cachable || (fetch_args$refresh_cache && is_cache_enabled())) {
```

tests/testthat/test-epidatacall.R
```diff
@@ -82,38 +78,7 @@ test_that("fetch_args", {
   )
 })
```
JavierMtzRdz, line 96: It may be worth keeping a test for the `fields` argument from the test below, since it is not tested elsewhere. 


```suggestion
test_that("fetch respects the fields parameter", {
  epidata_call <- pub_covidcast(
    source = "jhu-csse", signals = "sig", geo_type = "state", time_type = "day",
    geo_values = "ca", time_values = 20200101,
    fetch_args = fetch_args_list(dry_run = TRUE)
  )
  
  local_mocked_bindings(
    do_request = function(epidata_call, format_type, timeout_seconds, fields, ...) {
      # fields argument was passed down
      expect_equal(fields, "value")

      epidata_call <- extra_arguments(epidata_call, format_type, fields)

      # Verify that extra_arguments() appends fields to the URL
      expect_match(epidata_call$request$url, "fields=value")

      jsonlite::toJSON(list(
        epidata = list(list(value = 10)),
        result = 1,
        message = "success"
      ), auto_unbox = TRUE)
    },
    .package = "epidatr"
  )

  res <- fetch(epidata_call, fetch_args = fetch_args_list(fields = "value"))

  expect_equal(names(res), "value")
  expect_equal(res$value, 10)
})
```

R/epidatacall.R
```diff
@@ -188,8 +165,6 @@ fetch_args_list <- function(
   timeout_seconds = 15 * 60,
   base_url = NULL,
   dry_run = FALSE,
```
JavierMtzRdz, line 169: Unsure whether debug and format_type were used by anyone, but to be on the safe side, we could use a deprecation message. Note that if we use this, we need to add the package to DESCRIPTION.
```suggestion
  dry_run = FALSE,
  debug = lifecycle::deprecated(),
  format_type = lifecycle::deprecated(), 
```

JavierMtzRdz (approved):
The structure of the changes correctly consolidates multiple tasks. I included suggestions to fix typos and a possible bug in the cache. I also sent PR #337 to include additional cleanup of the tests and the suggested packages.

tests/testthat/test-epidatacall.R
dshemetov: This is definitely an improvement over the old test, thanks

R/request.R
dshemetov: Definitely much better. There's probably more of this sort of stopifnot-based validation in the rest of the package that could use this sort of update

R/epidatacall.R
dshemetov: Thanks for catching, I was too hasty here

R/epidatacall.R
dshemetov: Makes sense. (Side-note: might be worth doing the research required to figure out if we can just bump our requirement up to R >=4.1.0 and not leave many users behind).

R/epidatacall.R
dshemetov: Definitely reasonable, thanks

R/epidatacall.R
JavierMtzRdz: I saw that the Posit team [discussed](https://tidyverse.org/blog/2023/04/base-vs-magrittr-pipe/#using-the-native-pipe-in-packages) that, per their policy, they support R 3.5.0, and that examples and vignettes use the new pipe by conditionally evaluating the code chunks based on the version of R being used

dshemetov: I see, thanks for the link. I suppose it doesn't hurt us to just keep using the old pipe and support the older versions.

PR #343, opened 2026-05-18
doc: point to signal discovery for example queries everywhere

dshemetov:
Making the example query vignette more visible by linking to it.

### DIFF
--- a/R/covidcast.R
+++ b/R/covidcast.R
@@ -9,19 +9,9 @@ parse_signal <- function(signal, base_url) {
   class(signal) <- c("covidcast_data_signal", class(signal))
   signal$key <- paste(signal$source, signal$signal, sep = ":")
 
-  #' fetch covidcast data
-  #'
-  #' param data_source data source to fetch
-  #' param signals data source to fetch
-  #' param geo_type geo_type to fetch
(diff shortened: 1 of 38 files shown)


R/covidcast.R
```diff
@@ -9,19 +9,9 @@ parse_signal <- function(signal, base_url) {
...
-  #' param geo_values data source to fetch
-  #' param time_values data source to fetch
-  #' param as_of data source to fetch
-  #' param issues data source to fetch
-  #' param lag data source to fetch
-  #' return an instance of epidata_call
-  #' keywords internal
+  # Inner callable returned per signal: fetches covidcast data for the bound
+  # source/signal/time_type, taking geo_type, geo_values, time_values, and the
+  # versioning args (as_of, issues, lag). Returns an epidata_call.
```
dsweber2, line 14: should this also inherit from `.epidatr_shared_params`?

dsweber2 (approved):

R/covidcast.R
dshemetov: Pretty sure this is an internal function that roxygen doesn't pick up and therefore no docs generated

PR #352, opened 2026-07-15
Stream large API responses to disk to prevent OOM crashes

### LINKED ISSUE #304: Provide way to avoid OOM with county-level data access
This query
```r
dat <- epidatr::pub_covidcast("doctor-visits", "smoothed_adj_cli", "county", "day")
```
seems to produce a nondeterministic mix of problems:
- "Operation timed out after 900006 milliseconds with 0 bytes received" (on @XuedaShen's system)
- Raising an error because of low bytes/sec (on my system)
- (Potentially making the system unresponsive and then) getting killed by out-of-memory (OOM) killer (on my system).  I'm estimating (from state data * rough # counties per state) that data as an object would be ~0.7GB, and as a JSON string is around ~1.5GB; my system as ~32GB RAM, with 8.5GB "free" at the moment, so hypothetically it should be able to handle it in RAM without OOM killing, but disk might be an option too.

The last one seems like something epidatr could potentially help out with.

JavierMtzRdz:
### Checklist

Please:

      PR).
- [X] Makes sure to bump the version number in `DESCRIPTION`. Always increment
      number (the second number).

This pull request allows large API downloads using `fetch()` to succeed on memory-constrained machines without changing the returned tibble format.

Previously, `do_request()` loaded the entire response into memory, keeping duplicate copies of the raw data at once and causing out-of-memory crashes on large files. To fix this, `perform_and_read()`  now streams responses larger than 500MB to a temporary file instead of reading them directly into memory.

This PR added two new parameters to `fetch_args_list()`. `stream_threshold_bytes`  to set the streaming cutoff, and `download_path` to save the raw response to a specified file. If parsing into a tibble fails due to memory limits, this file is preserved so it can be queried using out-of-memory tools without downloading the data again.

- Resolves #304


R/epidatacall.R
```diff
@@ -352,6 +382,47 @@ request_epidata <- function(epidata_call, fetch_args = fetch_args_list(), simpli
   return(response_content$epidata)
 }
 
+# Read a response body with `reader`, whether it was kept in memory or streamed
+# to disk
+read_body <- function(res, reader, from_download_path) {
+  if (is.null(res$body_path)) {
+    con <- rawConnection(httr2::resp_body_raw(res))
+    # read_csv() closes the connection itself, but fromJSON()
```
dshemetov: Comment trails off. fromJSON() doesn't close the connection?

R/request.R
```diff
@@ -96,3 +108,88 @@ do_request <- function(epidata_call, format_type = c("json", "csv", "classic"),
...
+perform_and_read <- function(req,
+                             stream_threshold_bytes = Inf,
+                             download_path = NULL) {
+  resp <- httr2::req_perform_connection(req)
+  drained <- FALSE
+  # Close the response if it's not drained
+  on.exit(if (!drained) close(resp), add = TRUE)
+
+  status <- httr2::resp_status(resp)
+  content_length <- suppressWarnings(as.numeric(httr2::resp_header(resp, "Content-Length") %||% NA))
```
dshemetov: Have you tested this Content-Length property against our streaming server? I'm a bit skeptical that our API servers have this size information available (in some queries, the API server just forwards the rows along without knowing how many rows there will be in total). If our server always sends Content-Length NULL then we're always writing to disk by the logic here. If our server always sends a massive underestimate of Content-Length, we're always writing to memory and will get bit again.

So we might need a running tally of bytes received. Our approach then might need to change: either (a) hybrid: stream to memory, swap to file when threshold is exceeded, (b) simple: just always stream to a file.

dshemetov: One advantage of the "always file" approach is not having to manage byte chunks, `req_perform(req, path)` should just stream straight to a file. Might still be able to rely on that function in the hybrid approach, depending if req_perform can be made to append.

Drawback is the roundtrip to disk on all calls. We can try to measure it - iirc on many modern SSD machines, it's not a big penalty.

R/epidatacall.R
```diff
@@ -352,6 +382,47 @@ request_epidata <- function(epidata_call, fetch_args = fetch_args_list(), simpli
...
+    on.exit(try(close(con), silent = TRUE), add = TRUE)
+    return(reader(con))
+  }
+
+  path <- res$body_path
+  out <- tryCatch(
+    reader(path),
+    error = function(cnd) {
```
dshemetov: Is this error meant to trigger when the object is too big for memory? My guess is R won't be very graceful when that happens and the entire session will crash.

dshemetov (commented):
Might need to iterate on this one a couple times. Capturing OOM errors and streaming data is a bit tricky.

R/request.R
JavierMtzRdz: I tested it with some queries against the old API and V5. I noticed it was available inconsistently. Basically, when requesting an uncompressed transfer to the old API. That's why I set it to save it to file when it is NULL.
``` r
if (!requireNamespace("pacman", quietly = TRUE)) install.packages("pacman")
pacman::p_load(httr2, tibble, purrr)

cc <- "https://api.delphi.cmu.edu/epidata/covidcast/"
v5 <- "https://delphi.cmu.edu/epidata/v5/"

queries <- c(
  "cc 1 state, csv" = paste0(cc, "?data_source=jhu-csse&signal=confirmed_7dav_incidence_prop&time_type=day&geo_type=state&geo_values=ca&time_values=20200601-20200607&format=csv"),
  "cc 1 state, json" = paste0(cc, "?data_source=jhu-csse&signal=confirmed_7dav_incidence_prop&time_type=day&geo_type=state&geo_values=ca&time_values=20200601-20200607"),
  "cc all states, csv" = paste0(cc, "?data_source=jhu-csse&signal=confirmed_7dav_incidence_prop&time_type=day&geo_type=state&geo_values=*&time_values=20200601-20200630&format=csv"),
  "cc all counties, csv" = paste0(cc, "?data_source=jhu-csse&signal=confirmed_7dav_incidence_prop&time_type=day&geo_type=county&geo_values=*&time_values=20200601-20200607&format=csv"),
  "v5 snapshot, state" = paste0(v5, "snapshot/?source=nssp&signal=pct_ed_visits_combined&geo_type=state"),
  "v5 snapshot, county"  = paste0(v5, "snapshot/?source=nssp&signal=pct_ed_visits_combined&geo_type=county"),
  "v5 archive, state" = paste0(v5, "archive/?source=nssp&signal=pct_ed_visits_combined&geo_type=state")
)

# Status + Content-Length header from a streaming connection under one encoding.
peek <- function(url, gzip) {
  req <- request(url) |> req_error(is_error = function(resp) FALSE)
  req <- req_headers(req, "Accept-Encoding" = "identity")
  resp <- req_perform_connection(req)
  on.exit(close(resp))
  list(
    status = resp_status(resp),
    cl = resp_header(resp, "Content-Length") %||% NA_character_
  )
}

results <- imap(queries, function(url, nm) {
  g <- peek(url, gzip = TRUE)
  tibble(
    query = nm,
    status = g$status,
    cl = g$cl,
  )
}) |> list_rbind()

results
#> # A tibble: 7 × 3
#>   query                status cl     
#>   <chr>                 <int> <chr>  
#> 1 cc 1 state, csv         200 805    
#> 2 cc 1 state, json        200 2122   
#> 3 cc all states, csv      200 156573 
#> 4 cc all counties, csv    200 2209897
#> 5 v5 snapshot, state      200 <NA>   
#> 6 v5 snapshot, county     200 <NA>   
#> 7 v5 archive, state       200 <NA>
```
<sup>Created on 2026-07-23 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

JavierMtzRdz: I compared the median query times across different data sizes using three methods: in-memory gzip-compressed transfer, in-memory uncompressed transfer, and gzip transfer streamed straight to a temp file. On my machine, the in-file approach incurs overhead for small and medium queries. For that reason, the hybrid approach is appealing. 

Image

R/epidatacall.R
JavierMtzRdz: True. I just removed this.

JavierMtzRdz (commented):
I benchmarked different hybrid approaches, but as expected, there is no free lunch. As shown in the plot below, the hybrid approach seems to perform better on large queries, but for small and medium queries there is a clear overhead that is not much better than the always-file approach. 

For these tests, the hybrid threshold is 1 MiB. So, the hybrid approach saves the data in a file for all "large" queries. 

Based on some profiling tests, it appears that the main cause of increased time for small and medium queries is replacing `req_perform` with `req_perform_connection`. On the other hand, parsing is faster with the hybrid approach. Looking for alternatives to `req_perform_connection`.

Image

R/request.R
```diff
@@ -96,3 +107,78 @@ do_request <- function(epidata_call, format_type = c("json", "csv", "classic"),
...
+#' @param req an `httr2` request object
+#' @param stream_threshold_bytes see [do_request()]
+#' @param download_path see [do_request()]
+#' @return an `httr2_response`
+#' @keywords internal
+perform_and_read <- function(req,
+                             stream_threshold_bytes = Inf,
+                             download_path = NULL) {
+  resp <- httr2::req_perform_connection(req)
+  on.exit(close(resp), add = TRUE)
```
dshemetov, line 126: So this only covers connection close. We're still leaving half-transferred connections open if a (non-http) error throws (timeout, connection reset, disk full).

R/epidatacall.R
```diff
@@ -178,6 +185,8 @@ fetch_args_list <- function(
   disable_data_frame_parsing = FALSE,
   return_empty = FALSE,
   timeout_seconds = 15 * 60,
+  stream_threshold_bytes = 500 * 1024^2,
```
dshemetov, line 188: So I think peak memory use will be 3x stream_threshold_bytes - the  connection con, the chunks vector, and vctrs::list_unchop(chunks). At 3x 500MB, we're getting into 1.5GB. Might want to lower that. Also might be worth avoiding list_unchop and writing straight from the chunks vector, to reduce the memory peak factor.

R/request.R
```diff
@@ -96,3 +107,78 @@ do_request <- function(epidata_call, format_type = c("json", "csv", "classic"),
...
+      next
+    }
+    chunks[[length(chunks) + 1L]] <- chunk
+    # length() of a raw vector = number of bytes
+    bytes <- bytes + length(chunk)
+    if (bytes > stream_threshold_bytes) {
+      path <- tempfile()
+      cli::cli_inform(c("i" = "Streaming response to temporary file at {.file {path}}."))
+      con <- file(path, "wb")
+      writeBin(vctrs::list_unchop(chunks), con) # flush what we have so far
```
dshemetov, line 157: Avoids allocating more memory `for (ch in chunks) writeBin(ch, con)`. Idk if worth the speed penalty tradeoff.

tests/testthat/test-epidatacall.R
```diff
@@ -24,6 +24,58 @@ test_that("fetch surfaces http errors", {
...
+      function(...) {
+        if (!called) {
+          called <<- TRUE
+          raw(100)
+        } else {
+          raw(0)
+        }
+      }
+    }),
+    .package = "epidatr"
```
dshemetov, line 71: httr2::req_perform_connection is not actually being mocked here. You can test that with
```
  local_mocked_bindings(
    req_perform_connection = function(...) create_mock_response(""),
    resp_stream_raw = local({
      called <- FALSE
      function(...) {stop("MOCK USED")}
    }),
    .package = "epidatr"
  )
```

R/request.R
```diff
@@ -96,3 +107,78 @@ do_request <- function(epidata_call, format_type = c("json", "csv", "classic"),
...
+  if (httr2::resp_is_error(resp)) {
+    stream_threshold_bytes <- Inf
+    download_path <- NULL
+  }
+
+  # Holds chunks in memory until they exceed the threshold,
+  # then spilling to a file.
+  # `con` is non-NULL once we have switched to writing on disk.
+  chunks <- list()
+  bytes <- 0
```
dshemetov, line 138: I'd suggest calling this bytes_in_memory to be clear - we stop incrementing once the threshold is crossed.

dshemetov (commented):
Thanks for digging deeper into this stuff. A ~1s difference on the small requests seems like a pretty minor difference. My general outsider view is bias for the simpler thing, but the hybrid approach is not too complex, so I'll leave it up to you to make the call. Noted some small edge cases around errors and deferred on.exit() calls that need tweaks.

PR #353, opened 2026-07-15
Adds `epidata_aux()` for fetching V5-API auxiliary data

JavierMtzRdz:
### Checklist

Please:

      PR).
- [X] Makes sure to bump the version number in `DESCRIPTION`. Always increment
      number (the second number).

Adds `epidata_aux()` for fetching V5-API auxiliary data in two modes:

- By pulling `aux_data/` directly. Supports `filtered_keys`, `columns`, `report_time`, and `reference_time`. Following the approach in the previous `epidata_*` functions, it adds deprecated `time_values`/`issues`.
- By tibble from snapshot/archive output. It basically fetches the matching aux data and version-aware left-joins it onto the base. When `filtered_keys` isn't given, it's inferred from the base, so the pull stays small.
Also adds `fetch_args_list(disable_missing_meta_warning = FALSE)` to suppress unspecified fields warning to avoid declaring all columns in meta. 

In addition, it adds docs, a vignette section, and tests.

Some examples of how it works below: 

``` r
devtools::load_all(".")
#> ℹ Loading epidatr
#> Warning: package 'testthat' was built under R version 4.5.2
library(dplyr)
#> Warning: package 'dplyr' was built under R version 4.5.2
#> 
#> Attaching package: 'dplyr'
#> 
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> 
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union

nwss_data <- epidata_snapshot(
  source = "nwss",
  signal = "flu_avg_conc",
  geo_type = "sewershed",
  snapshot_date = "2026-07-06"
) |> 
  filter(sample_index == "92012")

head(nwss_data)
#> # A tibble: 1 × 10
#>   signal   report_time geo_type geo_value fill_method reference_time nwss_source
#>   <chr>    <date>      <chr>    <chr>     <chr>       <date>         <chr>      
#> 1 flu_avg… 2026-06-26  sewersh… 162       source      2026-01-27     CDC_Verily 
#> # ℹ 3 more variables: sample_index <chr>, pcr_target <chr>, value <dbl>

nwss_merged <- nwss_data %>% 
  epidata_aux()

head(nwss_merged)
#> # A tibble: 1 × 40
#>   signal   report_time geo_type geo_value fill_method reference_time nwss_source
#>   <chr>    <date>      <chr>    <chr>     <chr>       <date>         <chr>      
#> 1 flu_avg… 2026-06-26  sewersh… 162       source      2026-01-27     CDC_Verily 
#> # ℹ 33 more variables: sample_index <chr>, pcr_target <chr>, value <dbl>,
#> #   report_ts_nominal_end <chr>, state_territory <chr>, county_fips <chr>,
#> #   counties_served <chr>, population_served <chr>, sample_type <chr>,
#> #   sample_matrix <chr>, sample_location <chr>, flow_rate <chr>,
#> #   concentration_method <chr>, pasteurized <chr>, pcr_type <chr>,
#> #   extraction_method <chr>, major_lab_method <chr>, inhibition_detect <chr>,
#> #   inhibition_adjust <chr>, ntc_amplify <chr>, pcr_gene_target_agg <chr>, …

aux_data <- epidata_aux(
  source = "nwss",
  filtered_keys=c(
    pcr_target = "fluav",
    sample_index = "92012"
))

head(aux_data)
#> # A tibble: 6 × 36
#>   report_time geo_value reference_time nwss_source sample_index pcr_target
#>   <date>      <chr>     <date>         <chr>       <chr>        <chr>     
#> 1 2026-06-26  162       2026-01-27     CDC_Verily  92012        fluav     
#> 2 2026-06-19  162       2026-01-27     CDC_Verily  92012        fluav     
#> 3 2026-06-05  162       2026-01-27     CDC_Verily  92012        fluav     
#> 4 2026-05-30  162       2026-01-27     CDC_Verily  92012        fluav     
#> 5 2026-05-22  162       2026-01-27     CDC_Verily  92012        fluav     
#> 6 2026-05-15  162       2026-01-27     CDC_Verily  92012        fluav     
#> # ℹ 30 more variables: report_ts_nominal_end <chr>, state_territory <chr>,
#> #   county_fips <chr>, counties_served <chr>, population_served <chr>,
#> #   sample_type <chr>, sample_matrix <chr>, sample_location <chr>,
#> #   flow_rate <chr>, concentration_method <chr>, pasteurized <chr>,
#> #   pcr_type <chr>, extraction_method <chr>, major_lab_method <chr>,
#> #   inhibition_detect <chr>, inhibition_adjust <chr>, ntc_amplify <chr>,
#> #   pcr_gene_target_agg <chr>, pcr_target_units <chr>, lod_sewage <chr>, …
```

<sup>Created on 2026-07-15 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

- Resolves #{issue number}


R/endpoints.R
```diff
@@ -1418,6 +1420,226 @@ epidata_archive <- function(
...
+  aux <- epidata_aux(src, filtered_keys = filtered_keys, columns = columns, fetch_args = fetch_args)
+  if (!inherits(aux, "data.frame")) {
+    return(aux) # dry run: surface the aux call
+  }
+
+  # keep the newest aux version per key
+  if (ver %in% names(aux)) {
+    aux <- aux[order(aux[[ver]]), , drop = FALSE]
+    aux <- aux[!duplicated(aux[keys], fromLast = TRUE), , drop = FALSE]
+  }
```
dsweber2: So if I'm understanding this right, if we get a dataframe, we're attaching the most recent version of the auxiliary data? In the future, if given an archive this could be wrong, but it should be correct for nwss specifically

julesjulesjules: because we're not using report_time as a key anymore?

JavierMtzRdz: Yes, your interpretation is correct. Since the report times for both datasets are independent, I initially thought it would be easier to provide the latest versions for snapshots and archives, but that would mean merging each version in the base archive with the latest version of the auxiliary data for the same set of keys. I see how this approach can cause unexpected interpretation problems even with this dataset. I'll adjust it so the report time matches the specific version. It will be slightly more computationally expensive.

JavierMtzRdz: I end up using `vec_locate_matches` [[docs](https://vctrs.r-lib.org/reference/vec_locate_matches.html)]. This allows me to match each base row while finding the aux rows that match on every key and have a `report_time` at or before that base row's time. Among those, it keeps the one with the latest `report_time`. It ends up being ~2x faster for snapshots than the initial version.

R/endpoints.R
```diff
@@ -1418,6 +1421,242 @@ epidata_archive <- function(
...
+      # so a typed value (e.g. a Date inferred from the base) must go through
+      # `as.character` to serialize as "2024-01-01", not its integer day-count.
+      n <- vapply(filtered_keys, length, integer(1))
+      if (any(n != 1L)) {
+        cli::cli_abort(
+          "Each `filtered_keys` entry must have a single value; \\
+           {.field {names(filtered_keys)[n != 1L]}} {?has/have} more.",
+          class = "epidatr__epidata__multivalue_filtered_key"
+        )
+      }
```
dsweber2: We more or less talked about this, but it is actually supported on the backend to have multiple filters for each key, they just get passed in one at a time; we need to figure out good ways to support this. Whatever we choose we can more or less port it to `archive` and `snapshot` for free at least.

R/endpoints.R
```diff
@@ -1418,6 +1421,242 @@ epidata_archive <- function(
...
+    dropped <- if (!is.null(columns)) setdiff(keys, columns) else character()
+    if (length(dropped)) {
+      cli::cli_abort(
+        "`columns` excludes key column{?s} {.field {dropped}} needed to merge.",
+        class = "epidatr__epidata__missing_aux_keys"
+      )
+    }
+  }
+
+  # With no explicit `filtered_keys`, infer them from the base.
```
dsweber2: How to threshold this in a sane way when there's more than one value is a bit awkward; setting like 3k values in a URL may be a bad idea, but if the user has specified 10 it's probably worthwhile

R/endpoints.R
```diff
@@ -1418,6 +1421,242 @@ epidata_archive <- function(
...
+  # Match each base dataset row to the aux version current at its report_time
+  # with keys equal and aux report_time at or before the base's, keeping the newest
+  match_time <- if (identical(attr(base, "cast_kind"), "snapshot")) max(base[[ver]]) else base[[ver]]
+  m <- vctrs::vec_locate_matches(
+    needles = vctrs::vec_cbind(base[keys], .t = match_time),
+    haystack = vctrs::vec_cbind(aux[keys], .t = aux[[ver]]),
+    condition = c(rep("==", length(keys)), ">="),
+    filter = c(rep("none", length(keys)), "max"),
+    multiple = "any"
+  )
```
dsweber2, line 1667: @lcbrooks this is more or less the as_of logic from epiprocess no? Wondering if this function is newer than the implementation of `epix_as_of` or if you were avoiding it for specific reasons

tests/testthat/test-endpoints.R
```diff
@@ -278,3 +281,175 @@ test_that("epidata_archive local EpiRange filtering for report_time works", {
...
+  expect_match(call$request$url, "report_time_query=%3C2024-06-01") # "<" url-encoded
+  expect_match(call$request$url, "population_served") # columns
+  expect_match(call$request$url, "pcr_target") # filtered_keys serialized
+  expect_match(call$request$url, "2024-01-01") # Date value -> ISO, not day-count
+
+  # one value per key
+  expect_error(
+    epidata_aux("nwss", filtered_keys = list(geo_value = c("ca", "ny")), fetch_args = fetch_args_list(dry_run = TRUE)),
+    class = "epidatr__epidata__multivalue_filtered_key"
+  )
```
dsweber2, line 345: tests changes in to previous comments

dsweber2 (approved):
Should be fine to merge as-is if you want to defer work on the comments for later

R/endpoints.R
brookslogan: It didn't exist at the time I wrote the code we made into `epix_as_of`.  If it's on par with or faster than the `data.table` approach, it'd be nice to know; I think that would allow us another route other than duckplyr to jettisoning data.table data structures.  (I think I've tried dplyr rolling joins and they take twice the time; I assume they're probably backed by this vctrs function but maybe not / maybe they have a bunch of overhead.)

Side note: I don't understand most of what's going on here, so this may be completely misguided, but if this is doing an as-of operation, shouldn't `multiple` be something like ~~`"max"`~~ ["last" assuming sortedness; I guess "max" is not an actual option], not `"any"`?  Or maybe `"all"` since maybe `filter` being `"max"` will handle that part?

brookslogan: Looks like in the vctrs examples, they use the `multiple = "all"` default + the `filter` `"max"` approach:

```r
# From ?vctrs::vec_locate_matches

     # You can limit which matches are returned with a `filter`. For example,
     # with the above example you can filter the matches returned by `x[[i]] >= y`
     # down to only the ones containing the maximum `y` value of those matches.
     matches <- vec_locate_matches(x, y, condition = ">=", filter = "max")
```

JavierMtzRdz: Regarding the side note, we should not match multiple values. Maybe I should use `all` and return an error if the number of rows changes.

JavierMtzRdz: Yeah, `filter` should guard against multiple matches. No need to use "any."

brookslogan: I guess it depends on the order of `filter` and `multiple` whether it needs to be "all" or if it can just be anything.  If `filter` comes first, then `multiple` could be anything, and the non-"all" options might be more efficient as they could pre-construct a result vector of the appropriate length.  If `filter` comes last, then `multiple` has to be `"all"`.  I didn't see which one it was from skimming the docs.

brookslogan: Oh, I guess maybe they might have meant to imply the former by this; I originally read it as ambiguous:

> A filter can return multiple haystack matches for a particular needle if the maximum or minimum haystack value is duplicated in 'haystack'. These can be further controlled with 'multiple'.

(But hopefully this also means that they will also be doing whatever fast-preconstruction they might do with non-"all" even in the "all" case, since they know it won't be >1.  Not a guarantee though.)
[Actually, they don't, because they aren't equivalent:]

```r
x <- 1:2 + 0.5; y <- c(1,2,2,2,2,3) + 0.5; y[vctrs::vec_locate_matches(x, y, condition = ">=", filter = "max", multiple = "all")$haystack]
#> [1] 1.5 2.5 2.5 2.5 2.5
```

This suggests to go with your original approach "any", perhaps after seeing if there is any more confirmation that it's filter, then multiple.

PR #364, opened 2026-08-19
feat: error or warn on empty cast-API results (#356)

### LINKED ISSUE #356: Please raise error or warning in `epidata()` on some 0-row output cases
``` r
library(epidatr)
#> ! epidatr cache is being used (set env var EPIDATR_USE_CACHE=FALSE if not
#>   intended).
#> ℹ The cache directory is /home/fullname/.cache/R/epidatr.
#> ℹ The cache will be cleared after 14 days and will be pruned if it exceeds 4096
#>   MB.
#> ℹ The log of cache transactions is stored at
#>   /home/<username>/.cache/R/epidatr/logfile.txt.
# querying a non-existent geo_type simply yields 0 rows with no warning:
smoothed_pct_ed_visits_covid_not <- epidata("nssp", "smoothed_pct_ed_visits_covid", "hsa")
nrow(smoothed_pct_ed_visits_covid_not)
#> [1] 0
# vs. querying an actually-supported geo_type:
smoothed_pct_ed_visits_covid_hsa <- epidata("nssp", "smoothed_pct_ed_visits_covid", "hsa_nci")
nrow(smoothed_pct_ed_visits_covid_hsa)
#> [1] 139185
```

<sup>Created on 2026-08-14 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

It'd be nice if the first query could raise an error and list the correctly-"spelled", applicable geo types for the source.

However, some geo types are not comprehensively supported.  For example, some HSAs don't have NSSP data in CA (e.g., Alameda Contra Costa, which I believe should be 766).  That's not just a simple check against `epidata_meta("nssp")$nssp$geo_types`.  And one could get into the same situation when specifying time ranges as well.  I think we discussed what to do in these cases in v4, and it seems that we ended up keeping a warning when returning 0 rows?  That should be easy enough, though not as helpful as the much-more-complicated explanations of CA NSSP availability or outages /inactive systems.

dshemetov:
### Checklist

Please:

      PR).
      number (the second number).

- If the data comes back non-empty, but the local filters make it empty, give a warning.
- If the geo_type or signals are incorrect, give an explicit error

- Resolves #356


R/endpoints.R
```diff
@@ -1456,6 +1469,7 @@ epidata_archive <- function(
   attr(res, "cast_source") <- source # lets epidata_aux() recover the source
   attr(res, "cast_kind") <- "archive" # per-row revision history -> as-of aux merge
 
+  .check_cast_empty(res, fetched, source, signals, geo_type, fetch_args)
```
JavierMtzRdz, line 1484: suggestion: If the metadata is unlikely to fail, wouldn't it be better to move this as an initial validation of the query, before performing the call?

JavierMtzRdz (approved):
LGTM. I also included an optional suggestion.

R/endpoints.R
dshemetov: The approaches are basically the same in the failure case because empty-key queries are cheap on the db side. But validating lazily keeps the happy path zero cost (query is in local cache or query is already returning non-zero rows) and also prevents a stale/incomplete cache from blocking a request attempt.

PR #368, opened 2026-08-20
fix: bound vignette API queries with reference_time to avoid CI timeouts

dshemetov:
### Checklist

Please:

      PR).
      number (the second number).

Makes the API queries in the vignette smaller so CI doesn't timeout.


JavierMtzRdz (approved):
Since the filter is applied locally, I believe bounding `reference_time` won't affect the computation. Perhaps bounding it to an earlier `report_time` and `snapshot_date`. For example `report_time = "<2024-06-01"`.

dshemetov:
Ah good point, totally forgot about that. I suppose this becomes sorta moot with the next PR that takes the network call entirely out of CI (pre-compute vignettes locally).

PR #371, opened 2026-08-20
build: drop vignette-precompile packages from Suggests and usethis from Imports

dshemetov:
### Checklist

Please:

      PR).
      number (the second number).

Now that vignettes are precompiled, we can remove their dependencies from Suggests so R CMD check doesn't install them (this mostly speeds up ubuntu (devel) which doesn't get package binaries and has to compile them).

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -47,23 +47,19 @@ Imports:
     readr,
     rlang,
     tibble,
-    usethis,
+    utils,
     vctrs,
     lifecycle,
     xml2
 Suggests:
-    dplyr,
-    ggplot2,
     knitr,
-    mapproj,
-    maps,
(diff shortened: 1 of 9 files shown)


dsweber2 (approved):
It will make any attempts to recreate locally a tad more annoying, but I don't think that's terribly common anyway

DESCRIPTION
```diff
@@ -47,23 +47,19 @@ Imports:
     readr,
     rlang,
     tibble,
-    usethis,
+    utils,
```
JavierMtzRdz: question: do you see benefits for switching to utils directly? `usethis::edit_r_environ()` seems to offer robust project-root detection and IDE-aware file opening

